PR #16, opened 2023-07-20
Model simulation

martinjrobins:


pybop/models/spm.py
```diff
@@ -1,21 +1,12 @@
-import pybop
 import pybamm
 from .base_model import BaseModel
 
-class BaseSPM():
+class BaseSPM(pybamm.models.full_battery_models.lithium_ion.BasicSPM):
```
martinjrobins, line 4: There are two main methods to establish relationships between classes in OO, inheritance (which you are using), and composition. See here for more details: https://train.oxrse.uk/material/software_architecture_and_design/object_orientated/inheritance_and_composition_cpp

Composition tends to be more flexible as you don't lock yourself into an inflexible type hierarchy. This is especially disadvantageous if you are inheriting from a class you do not control.

So I would use composition here. Something like this:

```python
class SPM():
  def __init__(self):
    self.pybamm_model = pybamm.lithium_ion.SPM()
    self.name = "SPM"

File: pybop/simulation.py [hunk 2]
```diff
@@ -0,0 +1,23 @@
+import pybamm
+
+class Simulation(pybamm.simulation.Simulation):
```
martinjrobins, line 3: What are you thinking of doing with pybamm.Simulation? What functionality are you aiming to re-use?

martinjrobins: also, again I would use composition here rather than inheritance

martinjrobins (commented):
Hi @BradyPlanden, see comments below

pybop/models/spm.py
BradyPlanden: Excellent, thanks for the feedback @martinjrobins.

pybop/simulation.py
BradyPlanden: Initially, for pybop to run the pybamm forward model with the user needing to import pybamm. The longer-term aim is to diverge the parameterisation method so that it can be updated with needing to reconstruct the pybamm.Simulation class (unless there is a method to achieve this already?). [This](https://github.com/BradyPlanden/EstOptParticleModels/blob/6085049d323ba2ede34a79c1195349d71ebe881b/experiment/NLOpt-Fitting.py#L39) is how I've been doing it on a side project.

I wonder if inheritance would be best for this class if we plan to overwrite methods?

martinjrobins: ok I see. If you want to override methods than inheritance is the way to go. However, I would suggest that it might be good to write pybops own simulation class from scratch, since pybamm's simulation class makes a few assumptions that we do not want:
- it assumes a fixed set of parameters, we want a given set of input parameters
- it assumes a flexible experiment driven by conditions. We want a fixed experiment driven by measured data/timepoints

So our "experiment" and "simulation" classes would probably look a bit different.

martinjrobins: Note: In the linked python code above you contruct the simulation within the forward model, we definitely don't want to do this, there is a large setup cost for the models and solvers that we want to move outside the foward model. Particularly if you are using JAX

tests/Simulation.py
```diff
@@ -0,0 +1,29 @@
+import pybop
+import pybamm
+import numpy as np
+
+# Form list of applied current
+measured_expirement = [np.arange(0,3,0.1),np.ones([30])]
+current_interpolant = pybamm.Interpolant(measured_expirement[0], measured_expirement[1], pybamm.t)
```
BradyPlanden, line 7: Here is some sample usage with the current Simulation class, which is based on Pybamm.Simulation; however, it does not include the pybamm experiment methods as discussed above.

Question to the group, how far do we want to go in composing the pybamm methods? In this example, the only dependency on pybamm (at this level) is pybamm.Interpolant. Does it make sense for us to wrap the methods that are directly needed for pybop (creating drive cycles, updating parameters, plotting) and leave the rest for the user to acquire by importing pybamm?

BradyPlanden: To expand and add a specific example, we would build a pybop.Interpolant class from the `pybamm.Interpolant` as,

```python
class Interpolant():
  def __init__(self):
    self.Interpolant= pybamm.Interpolant()
```

This provides the user with a unified package without having a required dependancy on pybamm. This is different to how pybamm-param has been constructed [example](https://github.com/paramm-team/pybamm-param/blob/develop/examples/scripts/mle.py). However, we would need to build classes to compose for all possible pybamm functionality.

martinjrobins: I wouldn't restrict yourself to reproducing pybamm's api, feel free to make your own api that is more suitable for parameterised models. Then you can just re-use whatever pybamm functionality you want inside that api.

E.g.

```python
# Create abstract model (no data or parameters thus far)
model = pybop.BaseSPM()

# Create a parameterisation object (now include parameters to be inferred, experimental conditions like input current, and link data to model outputs)
observation = pybop.Observed(""Current function [A]", np.linspace(0, 1, 10))
parameters = [
   pybop.Parameter("Electrode height [m]", prior=pybop.Normal(0, 1))
   pybop.Parameter("Negative particle radius [m]", prior=pybop.Uniform(0, 1))
   pybop.Parameter("Positive particle radius [m]", prior=pybop.Uniform(0, 10))
]
parameterisation = pybop.Parameterisation(model, observations = [observation], parameters=parameter)

# get MAP estimate, starting at a random initial point in parameter space
parameterisation.map(x0=[p.sample() for p in parameters]) 

# or sample from posterior
paramterisation.sample(1000, n_chains=4, ....)
```

BradyPlanden:
Closing as this is superseded by #17.

Issue #18, opened 2023-08-04
How to handle structural parameters


BradyPlanden:
Parameters that don't affect the geometric structure of the model should be passed as inputs to the pybamm.solver class to improve performance. This will allow the built model to be updated with the appropriate parameters at each iteration without reforming the model, discretisation and solver.

If the parameters of interest have an influence on the geometric structure of the model (electrode thickness, number of particles, etc.), the above workflow must be rebuilt on each iteration. This slows down the iterations/sec of parameterisation/optimisation, but without pre-composing the design space (there is a side discussion to be had here) we need to do this to enable robust results. 

Open thoughts:

- For the structural parameter set, obtaining the gradient of the result will not be *easy*, and if needed will require more computationally complex methods, as the model structure changes with each iteration.
- There is a further discussion here about the fidelity of varying the model structure during either parameterisation or optimisation. There may be side effects that need to be captured in the model construction (for multi-particle models, the discretisation distribution could be held constant between structural changes).

To do:
- [ ] Group the parameters of supported models into the two domains of structural and non-structural.
- [ ] Update the constructor of the parameterisation/optimisation class to include functionality for each of the parameter groups.
- [ ] Build a parameter update method for each iteration that returns either the forward model or the simulation result to the parameterisation / optimisation method.

PR #20, opened 2023-08-16
Add testing framework and non-geometric parameterisation methods

### LINKED ISSUE #21: Add testing framework
Add an automated testing framework with Github actions. Tools to consider:

- Nox
- Tox

Needs to:

- [x] Have capability to test across python versions
- [x] Easy to create isolated testing enviroments 
- [x] Trigger on push, with additions to scheduled testing triggers at a later date
- [x] Have a low energy barrier to add new tests

BradyPlanden:
This PR adds the following:

- A testing framework built with [nox](https://nox.thea.codes/en/stable/index.html) & [pytest](https://docs.pytest.org/en/7.4.x/) for python versions 3.8 to 3.11
- Initial tests for the Parameterisation class
- Updates the Parameterisation class for non-geometric parameter optimisation as per #18
    - This builds the pybamm solver to include the fitting parameters as solver inputs. The next steps for this are outlined in #18, but include building methods to handle geometric parameters.
- Adds a parameter set class for application on construction of the parameterisation object 

Closes #21


examples/Initial_API.py
```diff
@@ -4,26 +4,26 @@
 
 # Form observations
 Measurements = pd.read_csv("examples/Chen_example.csv", comment='#').to_numpy() 
-observations = dict(
-    Time = pybop.Observed(["Time [s]"], Measurements[:,0]), 
-    Current = pybop.Observed(["Current function [A]"], Measurements[:,1]),
-    Voltage = pybop.Observed(["Voltage [V]"], Measurements[:,2])
-)
+observations = {
```
martinjrobins, line 7: Seems cleaner to have this as a list of `pybamm.Observed`, unless there is a reason to have it as a dict?

pybop/parameterisation.py
```diff
@@ -14,25 +13,48 @@ class Parameterisation:
...
+        geometry=None,
+        submesh_types=None,
+        var_pts=None,
+        spatial_methods=None,
+        solver=None,
+        x0=None,
+        check_model=True,
+        init_soc=None,
+    ):
+        self.model = model.pybamm_model
```
martinjrobins, line 30: seems confusing that you are storing the pybamm_model rather than the pybop one

martinjrobins: I would write `Parameterisation` assuming that you know nothing about pybamm. You have a pybop model which has a certain api, then just call functions on the pybop model (e.g. build_model()). The pybop model knows it stores a pybamm model, so any implementation details that depend on it being a pybamm model should go in the pybop model class

pybop/parameterisation.py
```diff
@@ -14,25 +13,48 @@ class Parameterisation:
...
-            upper=[p.bounds[1] for p in fit_parameters],
+            lower=[self.fit_parameters[p].bounds[0] for p in self.fit_parameters],
+            upper=[self.fit_parameters[p].bounds[1] for p in self.fit_parameters],
         )
-
-        if options is not None:
-            self.parameter_set = options.parameter_set
+        self._model_with_set_params = None
+        self._built_model = None
+        self._geometry = geometry or self.model.default_geometry
```
martinjrobins, line 40: again, a lot of pybamm specific stuff. Try to keep this all in a `BaseModel` class or something so if pybamm changes its api or we decide to start using other types of models, we only have to change a single class

pybop/parameterisation.py
```diff
@@ -41,15 +63,98 @@ def __init__(self, model, observations, fit_parameters, x0=None, options=None):
...
+        self.fit_dict = {
+            p: self.fit_parameters[p].prior.mean for p in self.fit_parameters
+        }
+
+        # Set t_eval
+        self.time_data = self.parameter_set["Current function [A]"].x[0]
+
+        self.build_model(check_model=check_model, init_soc=init_soc)
+
+    def build_model(self, check_model=True, init_soc=None):
```
martinjrobins, line 80: again, `build_model` should be in a model related class like `BaseModel`

pybop/parameterisation.py
```diff
@@ -41,15 +63,98 @@ def __init__(self, model, observations, fit_parameters, x0=None, options=None):
...
+        y_hat = self.solver.solve(
+            self._built_model, self.time_data, inputs=self.fit_dict
+        )[signal].data
+
+        try:
+            res = np.sqrt(
+                np.mean((self.observations["Voltage [V]"].data[1] - y_hat) ** 2)
+            )
+        except:
+            raise ValueError(
```
martinjrobins, line 150: I would directly check the length if you are going to raise this particular error, a ValueError could be a lot of other stuff too

tests/unit/test_parameterisation.py
```diff
@@ -0,0 +1,45 @@
+import pybop
+import pytest
+import numpy as np
+import pandas as pd
+
+
+class TestParameterisation:
+    def test_rmse(self):
+        # Form observations
+        Measurements = pd.read_csv("examples/Chen_example.csv", comment='#').to_numpy()
```
martinjrobins, line 10: using data files can be annoying if you try to run the tests from a different directory. Could we just generate some simulation data on the fly instead?

martinjrobins (approved):
Hi @BradyPlanden, this is a great start. i've made a bunch of suggestions, but at this early stage there is no need to fix them all before you merge, just something to keep in mind as you develop.

pybop/parameterisation.py
```diff
@@ -14,25 +13,48 @@ class Parameterisation:
...
+            self.parameter_set = model.parameter_set
         else:
-            self.parameter_set = self.model.pybamm_model.default_parameter_values
+            self.parameter_set = self.model.default_parameter_values
 
         try:
             self.parameter_set["Current function [A]"] = pybamm.Interpolant(
-                self.observations["Time"].data,
-                self.observations["Current"].data,
+                self.observations["Time [s]"].data,
```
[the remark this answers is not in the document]
BradyPlanden: Good shout, will add. 

This piece of code is the reason I have observations as a dictionary. It allows me to get the correct index based data without traversing a list, although I'm not very happy with how the dictionary is created in the API as you commented on above. 

Do you think it would be better to traverse the list and store the indexes as self variables in the constructor instead?

pybop/parameterisation.py
```diff
@@ -41,15 +63,98 @@ def __init__(self, model, observations, fit_parameters, x0=None, options=None):
         if x0 is None:
             self.x0 = np.mean([self.bounds["lower"], self.bounds["upper"]], axis=0)
```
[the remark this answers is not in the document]
BradyPlanden: Yeah absolutely, this needs to be updated to sample from the prior. I'll do that before merging.

tests/unit/test_parameterisation.py
BradyPlanden: Fair shout, this should be pretty easy to solve via importing pybamm and grabbing sample data from modified parameter simulation object.

pybop/parameterisation.py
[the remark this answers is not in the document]
martinjrobins: In the `__init__` I would just do:
```python
self.observations = { o.name: o for o in observations }
for name in ["Time [s]", "Current"]:
  if name not in self.observations:
     raise ValueError(f'expected {name} in list of observations')

```
where `observations` is the list of observations that are passed in

tests/unit/test_parameterisation.py
martinjrobins: try not to assume you are using pybamm too widely. simulating from a model will be a pretty common thing to do, so it might be useful to have the funcionality in the model class (or if it doesn't fit there a separate Simulation or Solver class)

Issue #6, opened 2023-07-14
Hamiltonian Monte Carlo (HMC)


BradyPlanden:
The [No-U-Turn ](https://arxiv.org/pdf/1111.4246v1.pdf) Hamiltonian Monte Carlo (HMC) method provide a robust method to capture a posterior distributions with efficient sampling and adaptive step size selection. Capturing the parameter posterior distributions for $\varepsilon = (\hat{y} - y$). This issue surmises the following:

- Integrate a NUTS implementation from a common, well-supported probablistic library
- Build corresponding test suite with code coverage
- Build diagonostics from posterior distribution (i.e. parameter observability) as well as performance-based metrics.

martinjrobins:
HMC (or NUTS is the varient always used in practice) is great. But for low numbers of parameters a simple method like Adaptive Covariance can be more reliable. Either way, loads of libraries give you NUTS and Adaptive Covariance (like PINTS ;) ) so I think getting both of these will be easy

PR #25, opened 2023-08-23
Working through a jax/numpyro implementation

BradyPlanden:
Using Numpyro and PyBaMM's JaxSolver class, we can create a just-in-time (jit) compiled NUTS sampler as per #6. Below is an example that aims to determine the best way to integrate these packages together. At the moment there are a few problems.

**Background**
PyBaMM's `JaxSolver()` has a method `create_solve` which creates a Jax function with a parameter dict as an input and the solution to the corresponding model's `rhs` as an output. To use this solution for the sampler, we need to acquire the model output variables ('Voltage [V]', 'Temperature [K]', etc.), which are usually constructed on the fly using the `ProcessedVariable` class.

**Issues**
One way to get the output variables is to construct the `pybamm.solution` and create the `ProcessedVariable` dict within the `mcmc.run` call. This would require the `pybamm.solution` class to be jit-capable (i.e. no casadi functions). This draft PR shows two attempts at this, which are flawed (read: don't work). To solve this, should we,

1. Create our own jit-capable solution class that converts the `rhs` solution returned by `create_solve` into the output variables.
2. Implement a callback (assuming we can do this in Jax) to the `pybamm.solution` class.
3. Add jit solution support to `pybamm.JaxSolver` (i.e. something like `pybamm.JaxSolver.solution`) so that `create_solve` returns the requested output variable(s).
4. Add jit support to the `pybamm.solution` class.
5. I'm missing something and there's already a nice solution available :)


martinjrobins:
So the sort of functionality (but for the idaklu solver) you are looking for is in the following issue: https://github.com/pybamm-team/PyBaMM/issues/2796 and pr https://github.com/pybamm-team/PyBaMM/pull/3217

Your main problem is that there is functionality in the solution class (and related classes) that executes the neccessary casadi code to convert the statevector vs time 2d array to the particular variables requested. Therefore you won't be able to jit this as it is based on casadi. 

You have to:
1. Convert the expression tree for each variable requested into a jax representation (perhaps using `pybamm.EvaluatorJax`)
2. use the pure jax function given by `pybamm.EvaluatorJax` in the Jax solver to convert the statevector to your output variables
3. Store these output variables *instead* of the statevector at each time point (again this is all in the solver)
4. return a lightweight solution class of your own making (i.e. jax-compatible) from the solver that only contains the output variables

In the PR linked above you can see how @jsbrittain has done this for the IDAKLU solver. One of the reasons for doing this in the first place was so that the solver is more suitable for use within a parameter estimation loop. So the same needs to be done for the Jax solver.

martinjrobins:
@jsbrittain is progressing on all this for the idaklu solver, but it will be a while before this is done and we can start working on the jax solver. Or if you want you could start on the jax solver straight away. Depends on how familar you want to get with the internals of Jax ;) (its valuable experience given that jax is so popular tho)

BradyPlanden:
Thanks for the feedback @martinjrobins, that's very helpful.

Right, that makes sense. Conceptually it's similar to option 3. above, with the difference being that we calculate and store the output variables in the state vector instead of using the `pybamm.solution` class as that is build upon Casadi. I'm happy to start working on the implementation for the JaxSolver (in a couple weeks after hols :grinning:).

PR #43, opened 2023-09-15
Update Architecture diagram

NicolaCourtier:


BradyPlanden:
On a whole, it looks great @NicolaCourtier! One small comment, the updated block named "Prior" was actually the non-optimised parameters required to run the forward model. 

To add a prior definition to this figure, I think it would need to be represented as an random draw from the the prior distribution during the initialisation of $\theta^*$ .

NicolaCourtier:
Thanks for the feedback @BradyPlanden! I have changed "Prior" to "Non-optimised Parameters" in line with your original design. We can consider how best to include a prior later.

Issue #50, opened 2023-09-26
Add forward simulation method


BradyPlanden:
### Feature description

Add a simple simulation class for synthetic data generation used in tests and benchmarking.

### Possible implementation

Built in `BaseModel`, this would provide functionality for the pybamm models; however, internal models such as a future EHM implementation would need their own definition of the `pybamm.Simulation` class. Given pybop follows the pybamm model class structure, it might be relavently simplistic to use the pybamm.Simulation class (will be brittle to pybamm changes though).
 
```python
 def sim(self, experiment=None, parameter_set=None):
        """
        Simulate the model
        """
        self.parameter_set = parameter_set or self.parameter_set
        if self._built_model is None:
            raise ValueError("Model must be built before simulating")
        return pybamm.Simulation(
            self._built_model,
            experiment=experiment,
            parameter_values=self.parameter_set
        )
```

Example usage:

```python
model = pybop.model.lithium_ion.spm()
synthetic_sim = model.sim()
term_V =synthetic_sim.solve()["Terminal voltage [V]"].data
```

PR #51, opened 2023-09-26
Update model structure, add sim() method and SPMe() class

BradyPlanden:
This PR updates PyBOP's model structure:

- `build()`, `set_init_soc()`, and `set_params()` methods have been moved to the `BaseModel` class
- `spm.py` is now `pybamm.py` with the addition of the `SPMe()` class. This givens an example on how to add pybamm models (further discussion on #7 is needed) . @NicolaCourtier, I think this is probably a halfway house between what we were discussing and what we currently have.  

Additionally, the BaseModel class adds a simulation method `sim()` for synthetic data generation of the corresponding pybop model (closing #50). This currently supports pybamm forward models, but future internal models should support the same functionality. 

Test cases for the `SPMe()` model class have been added.


BradyPlanden:
Merging after in person discussion :+1:

Issue #54, opened 2023-10-13
Code restructure for improved modularity


NicolaCourtier:
Restructure the code into component classes in which the cost function and optimisation method can be changed in isolation from one another.

PR #60, opened 2023-10-16
Restructure for improved modularity

NicolaCourtier:
Pull request to address issue #54


.gitattributes
BradyPlanden, line 1: Let's keep this

.gitignore
```diff
@@ -301,3 +301,6 @@ $RECYCLE.BIN/
 *.lnk
 
 # End of https://www.toptal.com/developers/gitignore/api/python,macos,windows,linux,c
+
+# Visual Studio Code settings
+.vscode/*
```
BradyPlanden, line 306: Let's keep

pybop/__init__.py
```diff
@@ -1,13 +1,13 @@
 #
-# Root of the pybop module.
+# Root of the PyBOP module.
```
BradyPlanden, line 2: We need to sort out a standard for this, real low priority though.

pybop/__init__.py
```diff
@@ -1,13 +1,13 @@
...
 # Provides access to all shared functionality (models, solvers, etc.).
 #
 # This file is adapted from Pints
 # (see https://github.com/pints-team/pints)
 #
 
 import sys
-import os
+from os import path
```
BradyPlanden, line 10: Maybe don't need to change, depends on whether we need any functionality from os in the future.

PR #62, opened 2023-10-23
47-add-self-hosted-runner

### LINKED ISSUE #47: Add self-hosted runner
### Feature description

Add the Battery Intelligence Lab's Apple M2 runner to PyBOP's CI for benchmarking and ARM validation.

### Motivation

As we expand PyBOP's estimation and optimisation methods, having a stable runner for benchmarking will provide a robust solution for new-users to compare and select methods.

### Possible implementation

The final solution should be *somwhat* similar to: https://github.com/pybamm-team/pybamm-bench with a similar representation deployed like: https://pybamm-team.github.io/pybamm-bench/

### Additional context

_No response_

BradyPlanden:
This PR adds a daily triggered unit test to be completed on the self-hosted M2 mini as well as the GitHub-hosted runners.

Note, the workflow names have been updated to align with their tasks. The dependency installations have also been refactored as there was redundant code. Closes #47

### DIFF
--- a/.github/workflows/scheduled_tests.yaml
+++ b/.github/workflows/scheduled_tests.yaml
@@ -1,0 +1,68 @@
+name: Scheduled
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+
+  # runs every day at 09:00 UTC
+  schedule:
+    - cron: '0 9 * * *'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


BradyPlanden:
This is now ready for review. A few items to note:

- nlopt.py's support for Apple silicon hardware is quite poor. I've had to build from source onto the runner, which is required for each python version we wish to test. This is quite combersome, and is not a pleasant workflow for macOS users. I'll open an issue to discuss this problem moving forward.
- At the moment, I've limited the runner to python 3.10. Once we sort out the above issue, this should be reverted to the original python matrix (as completed [here](https://github.com/pybop-team/PyBOP/blob/60d9a9e799c8cbd637a9130fa42a43298d01ac11/.github/workflows/scheduled_tests.yaml#L19))

martinjrobins (approved):
all looks great, thanks @BradyPlanden

PR #65, opened 2023-10-25
Run example scripts as tests

NicolaCourtier:
Move the long parameterisation tests into examples and then call all example scripts as tests.


BradyPlanden (approved):
Looks great! It was great to collaborate on this one 🥳

PR #73, opened 2023-10-27
Multi-parameter gradient information from PyBOP to Pints

BradyPlanden:
@martinjrobins would you mind taking a look at the implementation of the sensitivities in `simulateS1()` and the example `gradient_descent.py`? I've found that this will work for single parameter, but completely breaks down for multi-parameter. Pints is expecting an array of shape `(n_times, n_parameters)`, which I believe is sorted (although I'm not sure we need to convert from casadi.DM to numpy.. 

For more than a single parameter, the parameter estimation wanders throughout the space without converging. I suspect the gradient information isn't being correctly passed to Pints. The code is very brittle and hacked together at the moment as I'm working through understanding the interfaces (i.e. not to be merged).


BradyPlanden:
Closing this — solution found. Gradient information wasn't passed correctly when using the reshape() method. Returning the transposed gradient array to pints resulted in correct gradient based optimization.

Issue #82, opened 2023-11-03
Remove terminal voltage output constraint from simulate(), simulateS1()


BradyPlanden:
### Feature description

Currently, the BaseModel() class is constrained to output "Terminal Voltage [V]" in the simultate() and simulateS1() methods. This issue is to remove this constraint and add a signal variable to the BaseModel class.

### Possible implementation

self.signal = signal
...
return sol[signal].data

BradyPlanden:
Closed with #88

Issue #63, opened 2023-10-23
[Bug]: Nlopt install error for Apple silicon


BradyPlanden:
### Python Version

N/A

### Describe the bug

Installing nlopt.py on Apple silicon hardware results in a [build error](https://github.com/pybop-team/PyBOP/actions/runs/6611662098/job/17956004618) as shown in the log output below. Others have encountered this problem [[1]](https://github.com/DanielBok/nlopt-python/issues/13), and it appears that the current method of resolving it is to build nlopt from source. This is a tedious process that we shouldn't expect an end user to do. 

In this case, I think we have two options:

1.  Deprecate nlopt and replace it with scipy & pints
2. Have nlopt as an optional installation for Apple silicon hardware.

I believe they lead to roughly identical results, since it's clear that we can't rely on nlopt as a robust optimisation package in the future. My preference is option 1, but I am open to other ideas.

### Steps to reproduce the behaviour

`pip install nlopt` on Apple silicon hardware

### Relevant log output

```shell
Building wheels for collected packages: nlopt
  Building wheel for nlopt (setup.py): started
  Building wheel for nlopt (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [247 lines of output]
      running bdist_wheel
      running build
      running build_ext
      cmake version 3.27.7
      
      CMake suite maintained and supported by Kitware (kitware.com/cmake).
      CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
        Compatibility with CMake < 3.5 will be removed from a future version of
        CMake.
      
        Update the VERSION argument <min> value or use a ...<max> suffix to tell
        CMake that the project does not need compatibility with older versions.
      
      
      -- The C compiler identification is AppleClang 14.0.3.14030022
      -- The CXX compiler identification is AppleClang 14.0.3.14030022
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
[...]

PR #88, opened 2023-11-09
Integrate Pints optimisation methods

### LINKED ISSUE #72: Add Pints optimisation methods
### Feature description

This issue is to support adding [Pints ](https://github.com/pints-team/pints)methods into PyBOP. Pints has multiple entry points for this, including the [ForwardModel class](https://github.com/pints-team/pints/blob/5544f2b1f5b5cda4a646ddfae5a7aa2bf598aee3/pints/_core.py#L12), the [Problem](https://github.com/pints-team/pints/blob/5544f2b1f5b5cda4a646ddfae5a7aa2bf598aee3/pints/_core.py#L102) classes, and the [OptimisationController](https://github.com/pints-team/pints/blob/5544f2b1f5b5cda4a646ddfae5a7aa2bf598aee3/pints/_optimisers/__init__.py#L332) with the underlying `ask()` and `tell()` API. We already have support for the ForwardModel definition via our BaseModel class, with an MLE [example](https://github.com/pybop-team/PyBOP/blob/develop/examples/scripts/mle.py) showing how PyBOP and Pints can interact.

### Motivation

_No response_

### Possible implementation

_No response_

### Additional context

_No response_

BradyPlanden:
This PR completes the following:

**Optimisation**
- [x] Adds Pints GradientDescent() and CMAES() optimiser methods + example usage
- [x] Updates the Optimisation class for dual functionality (PyBOP optimisers & Pints optimisers) -> Credit to PINTS-team as a decent amount of the codebase is inherited from them
- [x] Updates optimiser argument to be class, and not object. Simplifies user API.

**Problem**

- [x] Renames the SingleOutputProblem class to Problem
- [x] Problem class now builds model

**Costs**

- [x] Updates the BaseCost class to compose the Problem class for clearer code divisions
- [x] Removes unused cost functions (should be readded at a later time)
- [x] Renames RootMeanSquareError() and adds SumofSquareErrors() 

**BaseModel**

- [x] Adds functionality for "Current function [A]" to be an input
- [x] Adds self.signal to replace hardcoded signal entries

**Tests**

- [x] Adds tests to get coverage passing
- [x] Updates current test suite for above changes

Closes #72, and #82,


examples/scripts/grad_descent.py
NicolaCourtier, line 1: Use corrupt_values instead of CorruptValues for variable name.

NicolaCourtier (approved):

NicolaCourtier:
Great work @BradyPlanden. I've approved the changes but wanted to make a couple of small comments:

1. Is there any reason not to merge ProblemCost with BaseCost? It seems like we should aim for one template here.
2. Much of `Optimisation` now seems to be specific to the PINTS framework. It may be worth extending the `Optimisation` class in a separate file until the capabilities of other optimisers have been aligned with the additional settings.

I've made these changes in #72b-add-pints-edits, so you can look and see if they help or not. Either way, I think it's good to merge, thanks Brady!

BradyPlanden:
Thanks for the review @NicolaCourtier!

1. Yes, nice catch. `ProblemCost` originally started as `SingleProblemCost` similar to Pints' implementation, but since we are only optimising single output models, this can be combined. Speaching of which, we should probably open an issue to add multi-output optimisation problems. 
2. I'm a bit hesitant to split the API of `Optimisation` at the top-level. I'm not quite happy with how `Optimisation` is at the moment, the logic conditions required to select between PyBOP optimisers and Pints' optimisers is not pretty. However, I think this does enable us to have one `Optimisation` class for the end-user to interact with. Perhaps we can do a better job of integrating the two structures to avoid this eyesore.. any ideas @martinjrobins?

NicolaCourtier:
1. Here's a possibly relevant open issue on Pints: https://github.com/pints-team/pints/issues/1372.
My opinion would be that we should try to avoid splitting the code along the single/multiple output dividing line.

2. Fair! I think it's a question of whether the additional settings will eventually apply to all optimisers or just the Pints optimisers. Now we have `Problem` as well as the `Optimisation` problem and the optimisers themselves, I am not sure where such settings belong.

NicolaCourtier:
I've been thinking about the big-picture purpose of each class and thought I would add some more thoughts here...

As far as I understand (please correct me @BradyPlanden !):

`BaseModel` and its extensions are classes which contain the capability to build a model including the (discretised) model equations, the list of fitting parameters (labelled as inputs), a parameter set, an (interpolated) input signal, and the name of the output signal. This class stores the dataset but only processes the time data and the input signal. A Model can be used to "simulate" the output given particular parameter values and time data, or "predict" the output corresponding to a PyBaMM Simulation based on an Experiment.

The `Problem` class stores the "built model" (after building it), the initial parameter estimates (x0), parameter bounds, the dataset and signal which define the time data and corresponding target data. Its main purpose seems to be to build the model. A Problem can be "evaluated" given particular parameter values, using the time data and simulate function of the built model.

`BaseCost` and its extensions are classes which can be "called" to "evaluate" the problem and return the value of the cost function given particular parameter values.

The `Optimisation` class builds the optimisation problem with a Cost (depending on a Problem and underlying Model), an optimiser and optimiser settings. A "run" of the Optimisation class returns best-fit estimates for the parameter values and the final cost.

The additional Pints definitions in Optimisation apply the ask/tell method of Pints as well as addtional optimiser settings.

`BaseOptimiser` and its extensions apply different optimisation alogrithms to the Optimisation problem specified by its cost, initial parameter values and parameter bounds.

So, my overarching questions would be: what are the advantages of having `Problem` in addition to and separate from the Model? And, is it necessary to implement the Pints optimisation framework inside `Optimisation` and not as an Optimiser interface?

BradyPlanden:
Hi @NicolaCourtier,

Yes, I think you've done a great job summarising the classes, thanks! A couple of small things:

- The `BaseModel` class probably doesn't need to store the dataset anymore (it currently stores a modified dataset structure). It will probably be more efficient to construct the interpolant function in the problem class and pass it to the model. The `model.predict()` method also allows predictions to be made without defining an experiment. We can actually clean up this class a bit more with the addition of the `Problem` class, specifically `model.n_outputs()` and `model.n_parameters()` can be removed.
- The `Problem` class stores the complete model object, including the built model as well as a gradient based evaluation method via `model.simulateS1`.

The `Problem` class gives us a couple of nice features. First, we have more control/security over the `model.build()` process (compared to a user building it) and a clearer error trace (compared to building in `Optimisation`, as we currently do). Second, conceptually I think it's nice to separate the preparation of the `BaseModel` class from the class itself, the build() dependancies is one such area, but another is sampling from the `Parameter' prior to update x0. 

On the Pints side, having it inside Optimisation() gives us a clean API to select optimiser methods (GradientDescent etc). When constructing this PR, I looked at keeping it inside the `PintsOptimiser` class as we currently do, the issue I had on this implementation was the API became:
```python
pybop.Optimisation(cost, optimiser=PintsOptimiser(pybop.GradientDescent))
```
or something similar, which I wanted to avoid. I don't think this implementation is the optimum solution, but it does provide a nice API for the end user with a start towards functionality that I think is helpful in the `Optimisation' class. There is also the #63 side of this discussion, where we need to consider whether NLOpt is something we want to keep long term. With that in mind, we might consider keeping the ask-tell interface as our base for the future. Let me know how this all jibes with you!

NicolaCourtier:
Thank you @BradyPlanden, I appreciate the detailed reply! It's really helpful to hear where you're heading with this, especially the better defined split between Model and Problem.

As I understand it now, rather than having one main `Optimisation` class which builds the optimisation problem from its components, we are migrating the problem definition to the sequential set-up provided by Model, Problem and Cost. Then we could either (as before) create an Optimiser and pass this to the `Optimisation` class, or allow `Optimisation` to build and call the Optimiser, analogous to the Pints `OptimisationController`.

I think it would be nice to allow users to call any Pints method and to avoid reimplementing them all by name, so I wondered if it might be possible to call Pints optimisers in a similar way to SciPy methods. For example, we could consider syntax such as:
`optimiser = pybop.PintsOptimiser(x0, method=pints.GradientDescent)`
`optim = pybopOptimisation(cost, optimiser=optimiser)`
I've had a play with this on #72b-add-pints-edits to see if/how it could work. The interesting part seems to be the choice of when and how optimiser options can be set.

BradyPlanden:
Hi @NicolaCourtier,

I've taken a look through the #72b branch. I have a few thoughts on these two differing approaches, I'll summarise them here,

1.  Using `pints.OptimisationController` in the PintsOptimisation class provides a clean, simplistic implementation that I also like. Although I think it's at the expensive of future flexibility, my concerns are,

       1. We will want more functionality than OptimisationController can provide quite quickly. Things like returning the optimisation state variables, or future modifications to optimiser structure that clash with Pints'. An example that comes to mind is returning the covariance matrix from CMAES to determine parameter sloppiness in the cost function.
       3. As you mentioned, the API for modifying the optimiser parameters becomes quite wieldy: `optim.optimiser.optim.max_iterations(300)` or we need to implement methods to modify the controller object (at certain times), which I think suffers from a similar problem to the one we are trying to avoid (reimplementing Pints methods in PyBOP).
       4. To avoid having Pints as a user included dependancy we would need to parse the method argument (`optimiser = "PintsGradientDescent"`) that corresponds to the class method and construct the Pints' object from that. Documenting this, while providing users with a clear knowledge on which methods are available might not be the cleanest.
   
Overall, I would prefer to maintain the Optimisation class with a similar OptimisationController codebase and work towards a genuine PyBOP implementation based on it. We would keep SciPy and NLOpt(?), and develop a tidier call structure for them in the Optimisation class. In the short-term, I think it is quite do-able to create more PyBOP inheritted classes from Pints' methods; I can probably do this before merging this PR. Let me know your thoughts on all of this!

NicolaCourtier:
Thanks @BradyPlanden!

Re. your first concern, it's not clear from #72b but I completely agree that we will develop the functionality of `pints.OptimisationController`. The difference is that #72b suggests putting the functionality of `pints.OptimisationController` in a separate subclass (`PintsOptimiser`) while #72 has it all in `Optimisation`. The underlying design decision is whether we are commiting to the Pints framework for all optimisation problems, or leaving the possibility of calling SciPy minimize (or others) directly without the setup for Pints appearing in this workflow. I believe your direction for #72 would be to wrap SciPy minimize inside of the Pints ask/tell framework, which would be another way to remove the PyBOP/Pints split in `Optimisation`. I worry that this could inadvertently encourage inefficient use of SciPy minimize but that it would work.

The implementation of optimiser options is currently similar in the 2 branches. The example `grad_descent.py` contains either:
[#72]   `optim.optimiser.set_learning_rate(0.025)` or
[#72b]  `optimiser.method.set_learning_rate(0.025)` followed by `optim = pybop.Optimisation(cost, optimiser=optimiser)`.
But you raise an interesting point about what happens when options are updated after `optim` has been built - I was assuming that a user would update the `optimiser` and then (re)create `optim`... triggering the whole setup to be run again, so users would likely opt to update `optim` directly and I agree that:
`optim.optimiser.method.set_learning_rate(0.025)` feels unnecessarily long! So let's just pass the name of the optimiser class to `Optimisation` as you suggest.

A compromise on the first point could be to have an `Optimisation` class that calls a `pints.OptimisationController`-type subclass while we work towards the tidier structure you have in mind? So that SciPy minimize remains an independent option for comparison purposes. Once we have tested SciPy minimize inside of the ask/tell framework then `Optimisation` would end up calling `OptimisationController` by default so the intermediary could be removed. We should update `Optimisation` to ensure that if no `problem` is contained within the `cost` argument then `optim` can still be created just with the `cost` function. This would make the `Optimisation` class a standalone module which may be tested independent of the full `problem`. We could even add a really simple `Cost` class e.g. `cost = abs(x)+42` for testing purposes.

Re. creating more PyBOP inherited classes, I think the end goal should be to allow users to call any PyBaMM or Pints method, so I'm not sure if this is needed - we should discuss with other user-developers, but it doesn't need to hold up the PR.

examples/scripts/rmse_estimation.py
```diff
@@ -30,31 +31,23 @@
...
-
-# or sample from posterior
-# parameterisation.sample(1000, n_chains=4, ....)
-
-# or SOBER
-# parameterisation.sober()
-
-
-# Optimisation = pybop.optimisation(model, cost=cost, parameters=parameters, observation=observation)
+x, final_cost = parameterisation.run()
```
martinjrobins, line 42: so I'm guessing that `run` returns a numpy array for `x`, and `problem.evaluate` takes a numpy array and returns a numpy array. I've never really liked this api for pints, because I felt as though there should be some labelling on which parameter was which (i.e. take and return a dict like object), and for the simulated values if you've got multiple outputs then you have to remember the right shape, what dimension is time and which is outputs. Plus the cost function can introduce multiple additional parameters which the user has not specified (e.g. sigma of noise). Just a comment for discussion, numpy arrays are fine, but perhaps we want to have something a bit more structured to add metadata like names.

BradyPlanden: Thanks for the review Martin. I agree, passing the numpy array back does lack information for the end-user. We should be able to update the parameter dictionary with the optimal value and return the entire dict. I'll add this as a future issue :)

Issue #98, opened 2023-11-17
Cost function returns an error when a simulation terminates early


NicolaCourtier:
If a PyBaMM simulation reaches a voltage constraint, the model can return a shorter solution array than expected, resulting in an error in the computation of a cost function such as RootMeanSquareError.

Optimisation problems are likely to run into such solutions when trialling different parameter values, but I think it should be taken to mean that the solution has a high cost rather than causing the optimisation to stop.

Suggested fixes:
1. Pad out any solutions that terminate early with the last output value to ensure that the prediction has the same length as the dataset.
2. Return a cost value of Infinity in the case that the prediction is shorter than the expected output.

I would opt for fix 2 for now, as I think it would give predictable behaviour and, in the case that all predictions are incompatible, it should return a final cost of Infinity.

BradyPlanden:
https://github.com/all-contributors, please add @NicolaCourtier to tests and examples.

BradyPlanden:
@all-contributors, please add @NicolaCourtier to tests and examples.

BradyPlanden:
@all-contributors
please add @NicolaCourtier for tests. 
please add @NicolaCourtier for examples.

BradyPlanden:
@all-contributors
please add @NicolaCourtier for test.

PR #99, opened 2023-11-17
Return Infinity if simulation terminates early

NicolaCourtier:
A resolution to issue #98.

### DIFF
--- a/pybop/costs/error_costs.py
+++ b/pybop/costs/error_costs.py
@@ -77,7 +77,7 @@ class SumSquaredError(BaseCost):
                     (np.sum(((prediction - self._target) ** 2), axis=0)),
                     axis=0,
                 )
-        
+
         except Exception as e:
             raise ValueError(f"Error in cost calculation: {e}")
 
--- a/pybop/costs/error_costs.py
+++ b/pybop/costs/error_costs.py
@@ -41,7 +41,7 @@ class RootMeanSquaredError(BaseCost):
             prediction = self.problem.evaluate(x)
 
             if len(prediction) < len(self._target):
-                return np.inf  # simulation stopped early
+                return np.float64(np.inf)  # simulation stopped early
             else:
                 return np.sqrt(np.mean((prediction - self._target) ** 2))
 
@@ -71,7 +71,7 @@ class SumSquaredError(BaseCost):
             prediction = self.problem.evaluate(x)
 
             if len(prediction) < len(self._target):
-                return np.inf  # simulation stopped early
+                return np.float64(np.inf)  # simulation stopped early
             else:
                 return np.sum(
                     (np.sum(((prediction - self._target) ** 2), axis=0)),
@@ -89,7 +89,7 @@ class SumSquaredError(BaseCost):
         try:
             y, dy = self.problem.evaluateS1(x)
             if len(y) < len(self._target):
-                e = np.inf
+                e = np.float64(np.inf)
                 de = self._de * np.ones(self.problem.n_parameters)
             else:
                 dy = dy.reshape(
--- a/pybop/costs/error_costs.py
+++ b/pybop/costs/error_costs.py
@@ -100,7 +100,7 @@ class SumSquaredError(BaseCost):
             )
             if len(y) < len(self._target):
                 e = np.inf
-                de = self._de*np.ones(self.problem.n_parameters)
+                de = self._de * np.ones(self.problem.n_parameters)
             else:
                 r = y - self._target
                 e = np.sum(np.sum(r**2, axis=0), axis=0)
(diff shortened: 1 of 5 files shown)


pybop/costs/error_costs.py
```diff
@@ -38,7 +38,14 @@ def __call__(self, x, grad=None):
         Computes the cost.
         """
         try:
-            return np.sqrt(np.mean((self.problem.evaluate(x) - self._target) ** 2))
+            prediction = self.problem.evaluate(x)
+
+            if len(prediction) < len(self._target):
+                cost = np.Infinity  # simulation stopped early
```
BradyPlanden, line 44: np.inf is probably a better constant to use: https://numpy.org/doc/stable/reference/constants.html

pybop/costs/error_costs.py
```diff
@@ -38,7 +38,14 @@ def __call__(self, x, grad=None):
         Computes the cost.
         """
         try:
-            return np.sqrt(np.mean((self.problem.evaluate(x) - self._target) ** 2))
+            prediction = self.problem.evaluate(x)
+
+            if len(prediction) < len(self._target):
```
BradyPlanden, line 43: We can return within the if statement:
```suggestion
            if len(prediction) < len(self._target):
                return np.Infinity  # simulation stopped early
            else:
                return p.sqrt(np.mean((prediction - self._target) ** 2))
```

pybop/costs/error_costs.py
```diff
@@ -57,10 +64,17 @@ def __call__(self, x, grad=None):
         Computes the cost.
         """
         try:
-            return np.sum(
-                (np.sum(((self.problem.evaluate(x) - self._target) ** 2), axis=0)),
-                axis=0,
-            )
+            prediction = self.problem.evaluate(x)
+
+            if len(prediction) < len(self._target):
```
BradyPlanden, line 69: Same as above

BradyPlanden (commented):
Nice one @NicolaCourtier! I've added a few comments.

Have you thought of any edge-cases on this? In certain design optimisation scenarios, it might be benefitical to return a smaller array (fast-charging comes to mind). There are other issues to solve before this would become the hold-up, but it's probably worth thinking of.

pybop/costs/error_costs.py
NicolaCourtier: Thanks for the link!

BradyPlanden:
This looks really good @NicolaCourtier. I think it actually solves #102 as well! I'd like to add a couple things on #102 to help the CMAES assertion, but they don't appear to be as needed. 

For testing this functionality, it would be good to add tests that push the optimiser outside of a stable solution. For reference, I've found setting negative active material fraction below ~0.4 in the SPM/e default parameter set can reliably crash the solver. See here:https://github.com/pybop-team/PyBOP/blob/2c7e74495425ba801aeb2b09670b55fa405465f3/tests/unit/test_parameterisations.py#L157

Happy for this to be merged once the coverage is passing 👍

NicolaCourtier:
I agree with your point above- there will be many edge cases where this is not the appropriate fix, but for the two Costs that we have now (RootMeanSquaredError and SumSquaredError) where there is a target to match, I think it gives predictable behaviour.

I think the downside would be that any dataset which spends time at eiher voltage limit could be very hard for the optimiser to find, because parameter sets that marginally break the limits will return an infinite cost even though they could be quite close to the optimal values. The first suggestion (extending the output array - there would be several choices we could make here) would give the optimiser more information in such scenarios and therefore possibly better performance in such edge cases.

pybop/costs/error_costs.py
```diff
@@ -38,7 +38,12 @@ def __call__(self, x, grad=None):
         Computes the cost.
         """
         try:
-            return np.sqrt(np.mean((self.problem.evaluate(x) - self._target) ** 2))
+            prediction = self.problem.evaluate(x)
+
+            if len(prediction) < len(self._target):
```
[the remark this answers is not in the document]
NicolaCourtier: I think we should leave this for now, we can expect to receive shorter solutions but I can't think of a reason for longer at the moment. I don't think that this fix should catch longer solutions as they would have an alternative explanation.

pybop/costs/error_costs.py
```diff
@@ -47,20 +52,32 @@ def __call__(self, x, grad=None):
...
         Computes the cost.
         """
         try:
-            return np.sum(
-                (np.sum(((self.problem.evaluate(x) - self._target) ** 2), axis=0)),
-                axis=0,
-            )
+            prediction = self.problem.evaluate(x)
+
+            if len(prediction) < len(self._target):
```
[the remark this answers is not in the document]
NicolaCourtier: Same as above

BradyPlanden (approved):

PR #105, opened 2023-11-17
102 - Update Pints' boundaries object

### LINKED ISSUE #102: [Bug]: Github pybop.CMAES tests are failing due to np.inf cost return
### Python Version

3.8-3.11

### Describe the bug

Failing tests here: https://github.com/pybop-team/PyBOP/actions/runs/6895515377/job/18759625531

Seperate from #98. I can't repeat the test fails locally (Intel Mac, AMD Ubuntu), but I suspect it's from the new optimiser options here:https://github.com/pybop-team/PyBOP/blob/2c7e74495425ba801aeb2b09670b55fa405465f3/tests/unit/test_parameterisations.py#L119. 

n/a

### Relevant log output

_No response_

BradyPlanden:
Fixes current failing tests. Closes #102.

Edit: This PR also,
-  Reduces the number of SOC locations that the unit tests complete, to improve GitHub actions runtime.
-  Replaces PintsBoundaries class with pint.RectangularBoundaries, which solves the randomly failing CMAES tests


pybop/optimisers/pints_optimisers.py
```diff
@@ -7,7 +7,7 @@ class GradientDescent(pints.GradientDescent):
     """
 
     def __init__(self, x0, sigma0=0.1, bounds=None):
-        boundaries = PintsBoundaries(bounds, x0)
+        boundaries = None  # Bounds ignored in pints.GradDesc
```
NicolaCourtier, line 10: Is it safe to overwrite the bounds input here? Should a message be returned to the user that bounds are being ignored?

pybop/parameters/base_parameter.py
NicolaCourtier, line 1: Should the lower < upper bound check allow lower *equal to* upper through?
If so, please update the message, otherwise use `>=`.

If they can be equal or less than 0.01% different then the adjustment of the samples to be within the bounds may fail and create values outside of the bounds.

I think this could be avoided and the adjustment and clipping performed together via e.g.
```
margin = 0.0001 * (self.upper_bound - self.lower_bound)
samples = np.clip(samples, self.lower_bound + margin, self.upper_bound - margin)
```

Do you think this would eliminate any possible unintended behaviour?

tests/unit/test_optimisation.py
```diff
@@ -92,6 +92,11 @@ def test_optimiser_construction(self):
             == "Covariance Matrix Adaptation Evolution Strategy (CMA-ES)"
         )
 
+        # None with no bounds
+        cost.bounds = None
+        opt = pybop.Optimisation(cost=cost)
```
NicolaCourtier, line 97: Should this check the CMAES optimiser explicitly? For GradientDescent, boundaries is None regardless of `cost.bounds` and the non-Pints optimisers do not have boundaries, is this correct?

NicolaCourtier (approved):
Well done for fixing the CMAES tests @BradyPlanden!
I left a few comments - please take or leave them as you wish.

pybop/parameters/base_parameter.py
BradyPlanden: Hmm, good question! 

Setting the margin from the range is a nice solution, although if lower == upper we will have a problem on the Pints side as per this [check](https://github.com/pints-team/pints/blob/220e49d9d304e847832810f10fc392e00a6e5c23/pints/_boundaries.py#L91). For the time being, I think I'll use the margin method and leave the lower < upper check. I'll also change the `0.0001` to be a hyper-parameter that can be user-modified.

tests/unit/test_optimisation.py
BradyPlanden: Yes, this will only check CMAES since it is our [default optimiser](https://github.com/pybop-team/PyBOP/blob/a66be81e409df321562c83d0df618e4821e9dbd2/pybop/optimisation.py#L51).

pybop/parameters/base_parameter.py
NicolaCourtier: Sorry, do you mean update the check so it is < or leave it as <=?

BradyPlanden: Updated to:

```python
if self.lower_bound >= self.upper_bound:
    raise ValueError("Lower bound must be less than upper bound")`
```

PR #115, opened 2023-11-21
Adds plotting methods

BradyPlanden:
This PR completes the following:

- [x] Adds a cost landscape (plot_cost2D) method
- [x] Default time-series plotting method for comparing
  - [x] Dataset
  - [x] Synthetic model with identified parameters
  - [ ] Slider functionality updating the fit for each optimisation step
- [x] Optimiser convergence plot method
- [x] Parameter value per iteration plot method

Update:
I've left the slider off this PR as future functionality to add. Additionally, the above plots work across all optimisers, except `plot_parameters()` and `plot_convergance()` which don't currently support NLopt and SciPy optimisers.


BradyPlanden:
This should now be ready for review. Four methods have been added:

1. `pybop.plot_2d(cost, optim=optim, steps=10)` which has an optional argument to overlay the optimiser evolution on the contour and # of steps.
2. `pybop.plot_convergence(cost)` plots the minimum cost value at each iteration of the optimisation.
3. `pybop.plot_parameters(cost)` create sub-plots for the evolution of the parameter values during the optimisation.
4. `pybop.quick_plot(cost)` creates a voltage against time figure for the identified parameters and the target data set.

`plot_convergence()` and `quick_plot()` both use the added StandardPlot class for figure creation. This class provide a standard figure implementation for single plot figures, this addition provides a base class for future figure expansion/additions. 

All the plotting is completed using plotly, with the setup.py updated to have an optional installation of `pip install pybop[plots]` or `pip install pybop[all]`.

See updated initial comment on functionality with nlopt and scipy.

Here are some sample figures, generated from the `spm_adam.py` example. First, `quick_plot()`:

newplot(1)

`plot_convergence()`:

newplot(2)

`plot_parameters()`:

newplot(3)

`plot_2d()` without optimiser evolution:

newplot(4)

and with (where the red marker is the starting position):

newplot(5)

conftest.py
```diff
@@ -1,6 +1,8 @@
 import pytest
 import matplotlib
+import plotly
```
NicolaCourtier, line 3: I see that you've added plotly as an extra requirement in the setup.py, but if it's not installed then this line returns an error, for example, with `pytest --unit -v`. Once installed of course, there is no issue. Should plotly be required here?

BradyPlanden: Good catch, the workflow for this is for developers to install the package with the [all] flag (i.e. `pip install -e '.[all]'`) which provides plotly. An update of the contributing.md is required to capture this change.

NicolaCourtier: We also recommend that new users run `pytest --unit -v` at the end of the Installation section of the Readme.

PR #125, opened 2023-11-28
Split Problem into Base, Fitting and Design

NicolaCourtier:
An initial PR to restructure Problem ahead of adding a design optimisation example.


BradyPlanden:
Hi @NicolaCourtier, nice one! How do you feel about changing `FittingProblem` to `FitProblem`?

pybop/_problem.py
```diff
@@ -1,30 +1,82 @@
...
+            self.x0 = np.zeros(self.n_parameters)
+            for i, param in enumerate(self.parameters):
+                self.x0[i] = param.rvs(1)
+        elif len(x0) != self.n_parameters:
+            raise ValueError("x0 dimensions do not match number of parameters")
+
+        # Add the initial values to the parameter definitions
+        for i, param in enumerate(self.parameters):
+            param.update(value=self.x0[i])
+        self.fit_parameters = {o.name: o.value for o in parameters}
```
BradyPlanden, line 42: We should rename this as it's current name doesn't make sense from the DesignProblem perspective. I think`self.parameters` works here since it end up as `pybop.FittingProblem.parameters` or `pybop.DesignProblem.parameters`

pybop/_problem.py
```diff
@@ -1,30 +1,82 @@
...
         parameters,
-        dataset,
-        signal="Terminal voltage [V]",
+        model=None,
         check_model=True,
         init_soc=None,
         x0=None,
     ):
         self._model = model
+        if model is not None:
```
BradyPlanden, line 18: I'm not sure this needs to be within a logic statement. If model = None, this still be calculated without issue, right?

pybop/_problem.py
```diff
@@ -1,30 +1,82 @@
...
+                self.x0[i] = param.rvs(1)
+        elif len(x0) != self.n_parameters:
+            raise ValueError("x0 dimensions do not match number of parameters")
+
+        # Add the initial values to the parameter definitions
+        for i, param in enumerate(self.parameters):
+            param.update(value=self.x0[i])
+        self.fit_parameters = {o.name: o.value for o in parameters}
+
+        # Then build the model
```
BradyPlanden, line 44: This can be removed

pybop/_problem.py
```diff
@@ -1,30 +1,82 @@
...
+class FittingProblem(BaseProblem):
+    """
+    Defines a PyBOP single output problem, follows the PINTS interface.
+    """
+
+    def __init__(
+        self,
+        parameters,
+        dataset,
+        model=None,
```
BradyPlanden, line 69: What happens if we try to use the FittingProblem without a model?

pybop/_problem.py
```diff
@@ -44,29 +96,62 @@ def __init__(
...
-                self.x0[i] = param.rvs(1)
-        elif len(x0) != self.n_parameters:
-            raise ValueError("x0 dimensions do not match number of parameters")
+        return (np.asarray(y), np.asarray(dy))
 
-        # Add the initial values to the parameter definitions
-        for i, param in enumerate(self.parameters):
-            param.update(value=self.x0[i])
+
+class DesignProblem(BaseProblem):
```
BradyPlanden, line 133: We need to update the docstrings for these new classes.

pybop/_problem.py
```diff
@@ -44,29 +96,62 @@ def __init__(
...
+        init_soc=None,
+        x0=None,
+    ):
+        super().__init__(parameters, model, check_model, init_soc, x0)
+        self.experiment = experiment
+        self._target = None
 
         # Set the fitting parameters and build the model
-        self.fit_parameters = {o.name: o.value for o in parameters}
         if self._model._built_model is None:
```
BradyPlanden, line 152: I think the logic for building the model should be added here instead of the BaseModel below. At this level we can add logic to trigger a build based on the value of `experiment`. This avoids modifying `build()`'s logic.

BradyPlanden:
It's looking good @NicolaCourtier! I've left few comments above, can you update the CHANGELOG.md as well?

pybop/_problem.py
NicolaCourtier: Good point. I think that it should be possible to use BaseProblem without a model (i.e. to allow for a StandaloneProblem), but FittingProblem and DesignProblem should probably require a model.

NicolaCourtier:
Thanks for the review @BradyPlanden! I've acted on your feedback so it would be great if you could take another look at the latest version.

examples/scripts/spm_CMAES.py
```diff
@@ -33,7 +33,7 @@
 ]
 
 # Generate problem, cost function, and optimisation class
-problem = pybop.Problem(model, parameters, dataset)
+problem = pybop.FittingProblem(parameters, dataset, model)
```
BradyPlanden, line 36: I think model could be the first parameter in the inputs, correct? This would allow us to avoid a breaking change here, and match user expectations coming from PyBaMM.

pybop/_problem.py
```diff
@@ -44,30 +93,67 @@ def __init__(
...
+        # Build the model if required
+        if experiment is not None:
+            # Leave the build until later to apply the experiment
+            self._model.parameters = self.parameters
+
+        elif self._model._built_model is None:
             self._model.build(
-                dataset=self._dataset,
-                fit_parameters=self.fit_parameters,
+                experiment=self.experiment,
```
BradyPlanden, line 155: Is experiment required in the model.build() method? Maybe for future functionality, perhaps?

tests/unit/test_cost.py
```diff
@@ -34,7 +34,9 @@ def test_costs(self, cut_off):
         # Construct Problem
         signal = "Voltage [V]"
         model.parameter_set.update({"Lower voltage cut-off [V]": cut_off})
-        problem = pybop.Problem(model, parameters, dataset, signal=signal, x0=x0)
+        problem = pybop.FittingProblem(
+            parameters, dataset, model=model, signal=signal, x0=x0
```
BradyPlanden, line 38: Since model is not optional, we can just provide model
```suggestion
            parameters, dataset, model, signal=signal, x0=x0
```

BradyPlanden (approved):
Nice one @NicolaCourtier! I left a few comments, but looks really good overall!

pybop/_problem.py
NicolaCourtier: Good spot - yes maybe but I'll leave it out for now.

PR #132, opened 2023-12-06
Add SciPyDifferentialEvolution optimiser

BradyPlanden:
This PR adds:

- `SciPyDifferentialEvolution` optimiser
- Adds `maxiter` as argument to BaseOptimiser with functionality added to `NLoptOptimize` and `SciPyMinimize`
- Adds the required tests for coverage


NicolaCourtier:
Hi @BradyPlanden, I added logging of the SciPy iterations to bring this branch up to speed with develop - this is the cause of the conflicts.

tests/unit/test_parameterisations.py
```diff
@@ -137,8 +138,13 @@ def test_spm_optimisers(self, spm_cost, x0):
                 x, final_cost = parameterisation.run()
 
             # Assertions
-            np.testing.assert_allclose(final_cost, 0, atol=1e-2)
-            np.testing.assert_allclose(x, x0, atol=1e-1)
+            # Note: SciPyMinimize has a different tolerance due to the local optimisation algorithms
+            if optimiser in [pybop.SciPyMinimize]:
+                np.testing.assert_allclose(final_cost, 0, atol=1e-2)
+                np.testing.assert_allclose(x, x0, atol=2e-1)
```
NicolaCourtier, line 144: When I see a failed assertion, it is usually the first one (final_cost) rather than the second (x).

I would prefer to avoid introducing a difference in the assertions here if possible.

BradyPlanden: Nice catch; however, while it fails on the cost assertion, the next failure would be on the "x" assertion. I'm not sure what choice we have, unless you've had success sorting the convergance issues with SciPyMinimize?

NicolaCourtier: I've made some progress, let's make it a different PR - thanks!

PR #134, opened 2023-12-06
Extend non-Pints optimisers for plotting

NicolaCourtier:
- Add logging for non-Pints optimisers
- Add a SciPy minimize example

### DIFF
--- a/examples/scripts/spm_scipymin.py
+++ b/examples/scripts/spm_scipymin.py
@@ -6,11 +6,11 @@ Measurements = pd.read_csv("examples/scripts/Chen_example.csv", comment="#").to_
 dataset = [
     pybop.Dataset("Time [s]", Measurements[:, 0]),
     pybop.Dataset("Current function [A]", Measurements[:, 1]),
-    pybop.Dataset("Terminal voltage [V]", Measurements[:, 2]),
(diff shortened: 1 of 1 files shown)


examples/scripts/spm_scipymin.py
```diff
@@ -0,0 +1,57 @@
...
+    ),
+]
+
+# Define the cost to optimise
+signal = "Terminal voltage [V]"
+problem = pybop.FittingProblem(model, parameters, dataset, signal=signal, init_soc=0.98)
+cost = pybop.RootMeanSquaredError(problem)
+
+# Build the optimisation problem
+optim = pybop.Optimisation(cost=cost, optimiser=pybop.SciPyMinimize)
```
BradyPlanden, line 38: I wonder if we should convert this to SciPyDifferentialEvolution once the corresponding PR is merged? As it's a global optimiser, we may want to recommend it more than SciPyMinimize. Thoughts?

BradyPlanden (approved):
Looks good @NicolaCourtier! One area that we might want to consider modifying for the future, otherwise LGTM

examples/scripts/spm_scipymin.py
NicolaCourtier: Sure, makes sense. I'll merge this one and then let's update in PR #132.

PR #146, opened 2023-12-13
Adds bayesian sampling

### LINKED ISSUE #6: Hamiltonian Monte Carlo (HMC)

BradyPlanden:
This PR adds Monte Carlo sampling methods; completed by extending [Pyro](https://pyro.ai/), and integrating the `pybop.problem` class.

To Do:
- [ ] Update `sigma` definition within `BayesianSampler.define_sampling_model` class
- [x] Integrate `parameter_prior` variable with `pybop.parameter`
- [ ] Add support for adding pyro dependencies after pip installation (similar to plotly)
- [ ] Add tests and notebook examples
- [ ] Update docs

Closes #6.


martinjrobins:
out of interest, why pyro? numpyro is faster and uses jax, so integrates well with pybamm. and PINTS has a lot of MCMC samplers (and slice samplers, SMC samplers....)

Nothing wrong with pyro of course!

BradyPlanden:
Hi @martinjrobins, great question - it comes down to the alignment between a Jax and a non-Jax implementation. From the Jax side, I agree that numpyro should be next on the list to be added, but it requires the completion of [#3395](https://github.com/pybamm-team/PyBaMM/issues/3395), which I'm hoping to complete over the holidays. 

At the top level, I think we should provide a Jax and non-Jax version of MCMC, as there are functional constraints on the Jax solver (e.g. I don't think it supports the experiment class, correct?). In this sense, using pyro and numpyro should allow us to reuse the design or at least minimise the structural changes. While using Pints' MCMC implementation would reduce the number of dependencies the project has, I think it would require two different approaches to MCMC between the JAX and non-JAX classes. I'm keen to hear your thoughts on this, while I have this initial Pyro example, I'm not too far into the implementation to pivot if we agree on a better design.

martinjrobins:
I'm not opposed to providing access to the pyro mcmc samplers, my main concern is that its not always possible to get gradients, or accurate gradients, and as far as I can see the only samplers in pyro are gradient-based ones (hmc and nuts). using pints would give you access to many more sampling methods, but its probably less optimised than pyro. 

Ideally, you would have three sampling backends, pyro, numpyro and pints.

PR #147, opened 2023-12-14
Reformat signal into a list

NicolaCourtier:
- Make `Dataset` contain a dictionary
- Extend `signal` into a list of strings
- Align NLopt and SciPy optimisers with `set_max_iterations`
- Parameterise `test_spm_optimisers` and add `test_multiple_signals`
- Update Changelog


pybop/models/base_model.py
```diff
@@ -204,15 +206,16 @@ def simulateS1(self, inputs, t_eval):
...
+            predictions = [sol[signal].data for signal in self.signal]
+
+            sensitivities = [
+                np.array(
+                    [[sol[signal].sensitivities[key]] for signal in self.signal]
+                ).reshape(len(sol[self.signal[0]].data), self.n_outputs)
+                for key in self.fit_keys
+            ]
+
+            return (np.vstack(predictions).T, np.dstack(sensitivities))
```
BradyPlanden, line 218: Are we wanting to return a tuple here?

pybop/optimisers/scipy_optimisers.py
```diff
@@ -62,6 +59,12 @@ def callback(x):
                 (lower, upper) for lower, upper in zip(bounds["lower"], bounds["upper"])
             )
 
+        # Set max iterations
+        if self._max_iterations is not None:
+            self.options = {"maxiter": self._max_iterations}
+        else:
+            self.options.pop("maxiter", None)
```
BradyPlanden, line 66: Nice one!

tests/unit/test_parameterisations.py
```diff
@@ -71,75 +71,108 @@ def spm_cost(self, parameters, model, x0):
...
+        else:
+            parameterisation.set_max_iterations(100)
+            x, final_cost = parameterisation.run()
+
+        # Assertions
+        np.testing.assert_allclose(final_cost, 0, atol=1e-2)
+        np.testing.assert_allclose(x, x0, atol=1e-1)
+
+    @pytest.fixture
+    def spm_cost_2(self, parameters, model, x0):
```
BradyPlanden, line 136: A docstring or a different name would make this fixture more clear. Something like `spm_two_signal_cost` would be good.

tests/unit/test_parameterisations.py
```diff
@@ -71,75 +71,108 @@ def spm_cost(self, parameters, model, x0):
...
+        signal = ["Terminal voltage [V]", "Time [s]"]
+        problem = pybop.FittingProblem(
+            model, parameters, dataset, signal=signal, init_soc=init_soc
+        )
+        return pybop.SumSquaredError(problem)
+
+    @pytest.mark.parametrize(
+        "optimiser",
+        [
+            pybop.NLoptOptimize,
```
BradyPlanden, line 158: Out of curiosity, what was the reason for these four optimisers?

pybop/plotting/quick_plot.py
```diff
@@ -215,21 +215,22 @@ def quick_plot(params, cost, title="Scatter Plot", width=1024, height=576):
...
-        yaxis_title=cost.problem.signal,
-        title=title,
-        trace_name="Model",
-        width=width,
-        height=height,
-    )()
-
-    # Display the figure
-    fig.show()
+    for i in range(0, cost.problem.n_outputs):
```
BradyPlanden, line 218: Longer term, we might want to consider creating a subplot for this. Have both signals on a single plot would be nice.

pybop/models/base_model.py
```diff
@@ -204,15 +206,16 @@ def simulateS1(self, inputs, t_eval):
...
-                        for key in self.fit_keys
-                    ]
-                ).T,
-            )
+            predictions = [sol[signal].data for signal in self.signal]
+
+            sensitivities = [
+                np.array(
+                    [[sol[signal].sensitivities[key]] for signal in self.signal]
+                ).reshape(len(sol[self.signal[0]].data), self.n_outputs)
```
BradyPlanden, line 214: Do we have an easier way to get `len(sol[self.signal[0].data])`? Maybe it can be taken from `self.target`? Looks good if not, just wondering if it could be simplified.

BradyPlanden (approved):
Looks great Nicola! I've added some comments to take a look at, hopefully they are helpful. Overall, great contribution  :+1:

tests/unit/test_parameterisations.py
NicolaCourtier: For speed of testing, I chose a subset covering the pybop options and the pints optimiser with and without sensitivities.
But I was wondering- which approach do you think we should take:
1. Test all optimisers using `spm_cost` and then test a subset again using `spm_two_signal_cost`, or
2. For speed, test half the optimisers with `spm_cost` and half with `spm_two_signal_cost`?

pybop/plotting/quick_plot.py
NicolaCourtier: I agree! Something to look into

pybop/models/base_model.py
NicolaCourtier: Yeah, I thought the same but ended up with this expression to ensure that it still works if the simulation terminates early. In this case, I think the length has to be taken from the newly-simulated `sol`.

tests/unit/test_parameterisations.py
BradyPlanden: Good shout, that makes sense to me too.

Hmm, I would consider test only a single optimiser for `spm_two_signal_cost`, or maybe just a single of each (pybop/pints). As an aside, what does the optimiser do when we have one of the signals as ["Time [s]"]? It's essentially a non-changing output, so maybe it doesn't change anything..

NicolaCourtier: Yeah, it shouldn't change the results, but including a second arbitrary signal here will flag if there is any mismatch in the array shapes. I think the set of 4 optimisers is required to properly check the array shapes are understood by all optimisers.

pybop/_problem.py
```diff
@@ -108,34 +108,44 @@ def __init__(
...
         if np.any(self._time_data < 0):
             raise ValueError("Times can not be negative.")
         if np.any(self._time_data[:-1] >= self._time_data[1:]):
             raise ValueError("Times must be increasing.")
 
-        if len(self._target) != len(self._time_data):
-            raise ValueError("Time data and signal data must be the same length.")
+        target = [self._dataset[signal] for signal in self.signal]
+        self._target = np.vstack(target).T
+        if self.n_outputs == 1:
```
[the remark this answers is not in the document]
NicolaCourtier: I like this shortening, but do you still prefer it with the formatting:
```
        if self.n_outputs == 1 and len(self._target) != self.n_time_data:
            raise ValueError("Time data and target data must be the same length.")
        elif self.n_outputs != 1 and self._target.shape != (
            self.n_time_data,
            self.n_outputs,
        ):
            raise ValueError("Time data and target data must be the same shape.")
```
?

pybop/_problem.py
```diff
@@ -108,34 +108,44 @@ def __init__(
...
-        self._target = self._dataset[signal].data
-
         if np.any(self._time_data < 0):
             raise ValueError("Times can not be negative.")
         if np.any(self._time_data[:-1] >= self._time_data[1:]):
             raise ValueError("Times must be increasing.")
 
-        if len(self._target) != len(self._time_data):
-            raise ValueError("Time data and signal data must be the same length.")
+        target = [self._dataset[signal] for signal in self.signal]
```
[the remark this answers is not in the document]
NicolaCourtier: Moved signal input to BaseProblem and added an error check.

pybop/models/base_model.py
NicolaCourtier: I believe this is what was returned before?

NicolaCourtier:
Thanks very much for the review @BradyPlanden! It would be great if you could double check the 3 conversations above and either add more info or mark as resolved.

pybop/models/base_model.py
BradyPlanden: Ah, right.
`(np.vstack(predictions).T, np.dstack(sensitivities)) == np.vstack(predictions).T, np.dstack(sensitivities)`

 It might be better to leave off the parentheses for consistency.

Issue #12, opened 2023-07-18
Kalman / covariance propogation solver


BradyPlanden:
This issue is to decide the implementation language for the solver required for #3.

Options discussed:
Pure Python
JAX
PyBaMM's IDAKLU (C++)

Currently, JAX is the method being considered as it provides compiled performance as well as code modularity between CPU/TPU/GPU.

davidhowey:
Just to drop in, as an aside, that I do quite like the idea of the "halfway house" observer we discussed a few days ago, i.e., fixed feedback gain parameter K that is learnt as part of the optimisation. Any observer is better than none, and this might mean we can use existing solvers.

martinjrobins:
is this still in scope for pybop? I notice that the `BaseModel` class currently handles the timestepping, i.e. timestepping, or the way time is discretised, is considered part of the model, and data only comes into the problem. But the output of a KF is (for parameter estimation anyway) the likelihood, which is a cost function, and the timestepping is also handled by the KF (or any observer), and data comes into the KF. So its unclear where a KF would fit into the current design.

It would be great if this would be included still!

BradyPlanden:
Hi Martin,

Yes, I think everyone is still keen for this to be included; but I agree that it's not clear how it fits into the current design.  I believe the fixed-gain observer could be completed by inserting a fixed gain optimisation parameter into the RHS (although for which state variable is the question). 

For a true KF implementation (and without thinking too much about it), we might be able to formulate the KF as a cost function with methods to change the time step by changing the problem variables/methods. Is this something that you are interested in looking at?

martinjrobins:
sure, I can look at this. I'm not exactly sure how to implement a KF for a general pybamm model, but I'll look into it....

PR #152, opened 2023-12-18
initial sketch of observers with UKF as method

martinjrobins:
Proposal for observers in PyBOP, using the unscented kalman filter as the observer method. See #12

I've added:
- `pybop.Observer`: a base observer class that just observes and does nothing else
- `pybop.UnscentedKalmanFilterObserver`: a derived observer class that uses the UKF to observe
- `pybop.ObserverCost`: a cost class that takes an observer and uses this to generate a cost function (takes the log likelihood of the observer and negates it)

A rather large disadvantage of the UKF for the SPM model is that it is very slow! I need to profile it and see where its taking all the time. Theoretically it should be 80 times slower, but it seems more than this.


BradyPlanden:
Hi @martinjrobins, this is looking great, thanks for the additions! I think it's proceeding in the right direction, shall I hold off on a review until your happy, or would like you like something at the current state?

Also, thanks for updating `BaseModel` with safety catches.

martinjrobins:
yea, don't review yet, doesn't even have proper tests and I'm sure things will change once I've done that.

martinjrobins:
ok, this is ready for review. I'm still sorting out some test failures on windows and python 3.8, but they are nothing to do with the observer class or the UKF.

btw. I'm not really sure what to do about the typing subscription error on python 3.8, any help appreciated!

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+        sqrtR : np.ndarray
+            The square root of the covariance matrix
+
+        Returns
+        -------
+        Tuple[np.ndarray, np.ndarray]
+            The mean and covariance of the sigma points
+        """
+        x = np.sum(w_m * sigma_points, axis=1).reshape(-1, 1)
+        A = np.sqrt(w_c[1:]) * (sigma_points[:, 1:] - x)
```
BradyPlanden, line 229: I think we could pre-calculate `sigma_points-x` to avoid calculating it a multiple points. 
```
sigma_points_diff = sigma_points - x
```

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+            The sigma points
+        sqrtR : np.ndarray
+            The square root of the covariance matrix
+
+        Returns
+        -------
+        Tuple[np.ndarray, np.ndarray]
+            The mean and covariance of the sigma points
+        """
+        x = np.sum(w_m * sigma_points, axis=1).reshape(-1, 1)
```
BradyPlanden, line 228: Can this be replaced with a dot product?

```
x = np.dot(w_m, sigma_points).reshape(-1,1)
```

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+        ) = linalg.qr(A.T, mode="economic")
+        S = UkfFilter.cholupdate(S, sigma_points[:, 0:1] - x, w_c[0])
+        return x, S
+
+    @staticmethod
+    def cholupdate(R: np.ndarray, x: np.ndarray, w: float) -> np.ndarray:
+        """
+        Updates the cholesky decomposition of a matrix (see https://github.com/modusdatascience/choldate/blob/master/choldate/_choldate.pyx)
+
+        Note: will be in scipy soon so replace with this: https://github.com/scipy/scipy/pull/16499
```
BradyPlanden, line 242: Excellent, we need to remember to update once that is merged & released.

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Returns
+        -------
+        float
+            The log likelihood of the measurement
+        """
+        sigma_points, w_m, w_c = self.gen_sigma_points(
+            self.x, self.S, self.alpha, self.beta
+        )
+        sigma_points = np.apply_along_axis(self.f, 0, sigma_points)
```
BradyPlanden, line 290: I wonder if we can vectorise `self.f` and `self.h` to avoid using `np.apply_along_axis`? https://stackoverflow.com/questions/23849097/numpy-np-apply-along-axis-function-speed-up

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+
+        x_minus, S_minus = self.unscented_transform(sigma_points, w_m, w_c, self.sqrtRp)
+        sigma_points_y = np.apply_along_axis(self.h, 0, sigma_points)
+        y_minus, S_y = self.unscented_transform(sigma_points_y, w_m, w_c, self.sqrtRm)
+        P = np.sum(
```
BradyPlanden, line 295: `np.einsum` might help with performance here. Something like:

```
P = np.einsum('i,ijk,ijl->kl', w_c, sigma_points - x_minus, sigma_points_y - y_minus)
```

pybop/observers/unscented_kalman.py
```diff
@@ -0,0 +1,309 @@
...
+            w_c * np.multiply.outer(sigma_points - x_minus, sigma_points_y - y_minus),
+            axis=(1, 3),
+        )
+        gain = linalg.lstsq(linalg.lstsq(P.T, S_y.transpose())[0].T, S_y)[0]
```
BradyPlanden, line 299: Did you try `linalg.solve` in-place of `linalg.lstsq`? Just wondering if there is stability issues with `linalg.solve` that requires us to use lstsq.

tests/unit/exponential_decay.py
```diff
@@ -0,0 +1,58 @@
+import pybamm
+
+from pybop.models.base_model import BaseModel
+
+
+class ExponentialDecay(BaseModel):
+    """
+    exponential decay model with two parameters y0 and k
```
BradyPlanden, line 8:
```suggestion
    Exponential decay model with two parameters y0 and k
```

tests/unit/test_observers.py
```diff
@@ -0,0 +1,42 @@
...
+import numpy as np
+import pybamm
+import pytest
+
+from tests.unit.exponential_decay import ExponentialDecay
+
+
+class TestObserver:
+    """
+    A class to test the unscented kalman filter.
```
BradyPlanden, line 11:
```suggestion
    A class to test the observer class.
```

BradyPlanden (approved):
Excellent work, @martinjrobins! I've added a few comments and suggestions, take what is helpful. If you can add an example notebook and update the CHANGELOG.md before merging that would be great.

pybop/models/base_model.py
```diff
@@ -131,10 +158,60 @@ def set_params(self):
...
+    def step(self, state: TimeSeriesState, time: np.ndarray) -> TimeSeriesState:
+        """
+        step forward in time from the given state until the given time.
+
+        Parameters
+        ----------
+        state : TimeSeriesState
+            The current state of the model
+        time : np.ndarray
+            The time to predict the system to
```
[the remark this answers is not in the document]
martinjrobins: I don't think it is necessarily seconds, its whatever units the model is using

pybop/models/base_model.py
```diff
@@ -131,10 +158,60 @@ def set_params(self):
...
+        """
+        dt = time - state.t
+        new_sol = self.solver.step(
+            state.sol, self.built_model, dt, npts=2, inputs=state.inputs, save=False
```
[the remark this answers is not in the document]
martinjrobins: I just wanted to be explicit so anyone reading the code would know without having to consult the docs for `step`

pybop/models/base_model.py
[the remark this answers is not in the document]
BradyPlanden: Gotcha, might be worth adding to the comment for future reference

pybop/observers/unscented_kalman.py
martinjrobins: perhaps? you are relying on `w_m` being broadcasted correctly which the `*` does atm. I can give `dot` a go and see if it works

pybop/observers/unscented_kalman.py
martinjrobins: its an over-determined set of equations, you are basically trying to fit a gaussian to a (possibly very) non-gaussian distribution, so `solve` won't work

pybop/observers/unscented_kalman.py
martinjrobins: I'd not used `einsum` before, looks good! I'll change this

pybop/observers/unscented_kalman.py
martinjrobins: Yea, this would have to be done somewhere in PyBaMM, needs to be a way to solve many identical pybamm models all starting from different initial conditions.

pybop/observers/observer.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Parameters
+        ----------
+        time : float
+          The time of the new observation.
+
+        value : np.ndarray (optional)
+          The new observation.
+        """
+        if time < self._state.t:
```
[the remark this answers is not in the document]
BradyPlanden: Nvm on this one, just noticed the line below.

pybop/models/base_model.py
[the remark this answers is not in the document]
martinjrobins: yup, added this to the docstring

pybop/observers/observer.py
```diff
@@ -0,0 +1,108 @@
...
+    ----------
+    model : BaseModel
+      The model to observe.
+    inputs: Dict[str, float]
+      The inputs to the model.
+    sigmal: List[str]
+      The signal to observe.
+    """
+
+    Covariance = np.ndarray
```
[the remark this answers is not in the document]
martinjrobins: I've added a comment to this effect

pybop/observers/unscented_kalman.py
martinjrobins: done

pybop/observers/unscented_kalman.py
martinjrobins: see https://github.com/pybamm-team/PyBaMM/issues/3713

pybop/observers/unscented_kalman.py
martinjrobins: @BradyPlanden: I tried and failed at this :( the string `i, ijk, ijl` does not work because sigma_points - x_minus is only a 2D array, not 3D. I couldn't come up with the right string to first do the outer product, and then the sub over axes 1 + 3. Any ideas?

martinjrobins: actually this seems to work with the sizes, but gives a different answer than the previous implementation!

```python
P = np.einsum("k,jk,lk -> jl ", w_c, sigma_points - x_minus, sigma_points_y - y_minus)
```

martinjrobins: tests still pass though! I think I need a better test on the covariance calculation.....

martinjrobins: actually the new einsum implementation is the correct answer :) fixed now

martinjrobins: still needs some better testing. Perhaps after this is merged in I can implement a kalman filter as well. For a linear model the kalman filter and the ukf should give identical answers correct?

martinjrobins:
@BradyPlanden, what are you thinking on python 3.8 support. I've used type subscripting a lot which 3.8 does not support. Should I get rid of the type subscripting or should we drop 3.8 support?

martinjrobins:
> @BradyPlanden, what are you thinking on python 3.8 support. I've used type subscripting a lot which 3.8 does not support. Should I get rid of the type subscripting or should we drop 3.8 support?

never mind, I think we can just do `from __future__ import annotations`!

BradyPlanden:
Excellent, it would be great to maintain python 3.8 support until PyBaMM drops it, which, for visibility is being discussed in [this](https://github.com/pybamm-team/PyBaMM/issues/3390) issue.

pybop/observers/unscented_kalman.py
martinjrobins: nope, `dot` does not do the broadcasting :( 

```
ValueError: shapes (2,) and (4,2) not aligned: 2 (dim 0) != 4 (dim 0)
```

BradyPlanden: Makes sense, thought it was worth a double-check. Thanks for investigating

martinjrobins:
any ideas on the example script failure? I tried on my machine and it worked fine, doesn't seem to be anything to do the observer or ukf

BradyPlanden:
I'm not sure why we are getting that error, it seems to be linked to convergence problems when using a fake cost/gradient in infeasible locations. Shouldn't be caused by the observer, and I think it's solved in #166.

NicolaCourtier (approved):
Thanks very much @martinjrobins! Looks good to me, please merge when you are happy.

BradyPlanden:
@all-contributors please add @martinjrobins for code and tests

PR #153, opened 2023-12-18
Add StandaloneProblem

NicolaCourtier:
- Add StandaloneProblem to examples
- Move StandaloneCost in same folder
- Add test for Standalon Problem


BradyPlanden (approved):
LGTM, nice work @NicolaCourtier. 

Looking into the future, I think we will want to provide a nicer interface to construct standalone problem and cost classes.

*Edit:* Can you update the Changelog as well?

PR #166, opened 2024-01-10
Revert `np.inf` return for non-physical optimisation locations

### LINKED ISSUE #164: Adam optimiser failing tests
### Python Version

N/A

### Describe the bug

Convergence issues for the Adam optimiser. Potential areas to investigate:

- [x] The solvers response to `np.inf`
- [ ] Additional iterations required?

`nox -s unit`

### Relevant log output

```shell
https://github.com/pybop-team/PyBOP/actions/runs/7464440347/job/20311432705
```

BradyPlanden:
When returning `np.inf` for non-physical optimisation iterations, the gradient-based optimisers struggle to converge due to the lack of stable gradient information (either with a fixed value or trying to create a fake gradient). Currently, for the active material fitting examples & tests, the `np.inf` occurs infinitesimally past the cost minimum. This is a difficult challenge to solve robustly for different optimisation cases, so instead of trying to integrate the fake gradient/cost into the landscape, I suggest we leave the default functionality to not return `np.inf` and replace it with a `UserWarning`.

To do this I've add the argument `allow_infeasible_solutions` to `model.check_params()` that modifies that silences the method return (i.e return `True`). I've also added a `UserWarning` with a counter to limit the warning displayed during optimisation. This allows the cost and gradient surfaces to remain continuous and has largely resolved the test suite failures, but still retains the original functionality if silent passed as `True`

In addition, the gradient-based optimisers are quite sensitive to initial conditions, so I initialised them at a lower gradient location to help with convergence. This is probably something we should think about and update in the longer term. At the very least we should keep it in mind when recommending optimisers. This PR also increases the `max_iterations` to help reduce the number of sporadic convergence failures we experience.

Closes #164.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Bug Fixes
 
+- [#164](https://github.com/pybop-team/PyBOP/issues/164) - Fixes convergence issues with gradient-based optimisers, changes default `model.check_params()` to allow infeasible solutions during optimisation iterations. Adds a feasibility check on the optimal parameters.
+
 # [v23.12](https://github.com/pybop-team/PyBOP/tree/v23.12) - 2023-12-19
 ## Features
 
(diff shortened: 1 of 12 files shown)


NicolaCourtier (approved):
Thanks @BradyPlanden! I've updated the design problem so it does not allow `infeasible_locations` and the descriptions. Ready to merge!

PR #169, opened 2024-01-18
Add fitting functionality for geometric parameters

### LINKED ISSUE #18: How to handle structural parameters

BradyPlanden:
This PR adds functionality to fit geometric parameters in pybop. This essentially requires a model rebuild every iteration, but a combined effort to minimise the number of objects that get rebuilt. It completes the following:

- [ ] Adds `rebuild()` method to `base_model`
- [ ] Adds functionality to classify and store parameters based on their type (requires rebuild or not) via `model.rebuild_parameters`
- [ ] Adds `set_rebuild_parameters()` to create the required dictionary of geometric parameters
- [ ] Adds a `copy()` method to `base_model`
- [ ] A type change to `plot_cost2D()`
- [ ] Adds tests needed for coverage

Closes #18

### DIFF
--- a/examples/scripts/spm_geometric_parameters.py
+++ b/examples/scripts/spm_geometric_parameters.py
@@ -1,0 +1,60 @@
+import pybop
+import numpy as np
+
+# Define model
+parameter_set = pybop.ParameterSet.pybamm("Chen2020")
+model = pybop.lithium_ion.SPM(parameter_set=parameter_set)
+
+# Fitting parameters
+parameters = [
+    pybop.Parameter(
+        "Positive particle radius [m]",
+        prior=pybop.Gaussian(5.22e-06, 0.05e-06),
+        bounds=[4e-06, 6e-06],
+    ),
+    pybop.Parameter(
+        "Positive electrode active material volume fraction",
+        prior=pybop.Gaussian(0.48, 0.05),
+        bounds=[0.4, 0.7],
+    ),
+]
+
+# Generate data
+sigma = 0.001
+t_eval = np.arange(0, 900, 2)
+values = model.predict(t_eval=t_eval)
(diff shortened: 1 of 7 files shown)


BradyPlanden:
Hi @NicolaCourtier, this should now be ready for review. I've updated the spm_CMAES.py example to be geometric as a reference. Seems to be working fine, and isn't a massive hit on performance.

NicolaCourtier:
Did you mean to change `spm_CMAES` as well as add `spm_geometric_parameters` as a separate example? They look very similar to me. Great to hear that it's up to speed!

NicolaCourtier (approved):
Looks good to me, thanks @BradyPlanden, don't forget the Changelog!

PR #178, opened 2024-01-29
Update plotting capabilities and dataset restructure

NicolaCourtier:
* Update plotting, descriptions and examples - the same plots are implemented using a more flexible design based on StandardPlot - for example, layout keyword arguments can be passed rather than individual properties such as `xaxis_title`
* Add __getitem__ and check to Dataset class
* Add plot_dataset


BradyPlanden:
Hi Nicola,

Thanks for the addition! Unforunately, I'm a bit stuck, as I'm not sure how to proceed with the current state of the diffs. I don't think we will be able catch differences in the base vs head due to the full file deltas. I'm not very keen on manualing diffing them, perhaps we can remove the file name changes, and add them after the PR is approved. 

Could you also provide details on the codebase changes in the top PR comment? It would both help me and give us a bread trail for changes on these files.

NicolaCourtier:
I've merged from develop and reverted the filenames, although at least one plotting script has moved to a different file from where it was before. Are you able to review the changes in this state or do the changes need to be further broken down?

pybop/plotting/plot_cost2d.py
```diff
@@ -48,112 +52,51 @@ def plot_cost2d(cost, bounds=None, optim=None, steps=10):
         for j, yj in enumerate(y):
             costs[j, i] = cost([xi, yj])
 
-    # Create figure
-    fig = create_figure(x, y, costs, bounds, cost.problem.parameters, optim)
+    # Import plotly only when needed
+    import plotly.graph_objects as go
```
BradyPlanden, line 56:
```suggestion
    go = pybop.PlotlyManager().go
```

BradyPlanden (commented):
Hi @NicolaCourtier,

Thanks for this addition, it adds a lot of good stuff to the plotting classes! I've added a few comments, but I've also split off a branch so as not to confuse this PR too much. 

I think the only larger discussion we should have is around `plot_optim2d` and `plot_cost2d`. My preference on this would be to merge them like develop currently has, and use an argument for the user to add the optim trace. I like passing `optim` instead of `cost` as you've done here, but I think a transition to something more general like `plot_2d` with arguments or adding logic based on the input variable would be ideal. This would reduce the number of places where we have duplicate code, and the number of different plot functions that end users would need to know.

NicolaCourtier:
Hey @BradyPlanden, thanks for creating a branch to develop your suggestions - it's looking great at the moment!

pybop/plotting/plot2d.py
```diff
@@ -0,0 +1,203 @@
+import sys
+import pybop
+import numpy as np
+
+
+def plot2d(
+    cost_or_optim, gradient=False, bounds=None, steps=10, show=True, **layout_kwargs
```
BradyPlanden, line 7: Something like this could be a bit more clear. Could helpful with generality in the future as well.

```suggestion
    plot_object, gradient=False, bounds=None, steps=10, show=True, **layout_kwargs
```

pybop/plotting/plot2d.py
```diff
@@ -0,0 +1,203 @@
...
+):
+    """
+    Plot a 2D visualisation of a cost landscape using Plotly.
+
+    This function generates a contour plot representing the cost landscape for a provided
+    callable cost function over a grid of parameter values within the specified bounds.
+
+    Parameters
+    ----------
+    cost_or_optim : a callable cost function, pybop Cost or Optimisation object
```
BradyPlanden, line 17:
```suggestion
    plot_object : a callable object that returns a scalar value. Example: pybop.cost or pybop.optimisation
```

pybop/plotting/plot2d.py
```diff
@@ -0,0 +1,203 @@
...
+        The Plotly figure object containing the cost landscape plot.
+
+    Raises
+    ------
+    ValueError
+        If the cost function does not return a valid cost when called with a parameter list.
+    """
+
+    # Assign input as a cost or optimisation object
+    if isinstance(cost_or_optim, pybop.Optimisation):
```
BradyPlanden, line 46:
```suggestion
    if isinstance(plot_object, pybop.Optimisation):
```

pybop/plotting/plot2d.py
```diff
@@ -0,0 +1,203 @@
...
+        optim = cost_or_optim
```
BradyPlanden, line 47:
```suggestion
        optim = plot_object
```

pybop/plotting/plot2d.py
```diff
@@ -0,0 +1,203 @@
...
+        plot_optim = True
+        cost = optim.cost
+    else:
+        cost = cost_or_optim
```
BradyPlanden, line 51:
```suggestion
        cost = plot_object
```

tests/unit/test_plots.py
```diff
@@ -91,4 +119,15 @@ def test_optim_plots(self, optim):
...
 
         # Plot the cost landscape with optimisation path
-        pybop.plot_cost2d(optim.cost, optim=optim, steps=5)
+        pybop.plot2d(optim, steps=5)
+
+    @pytest.mark.unit
+    def test_with_ipykernel(self, dataset, cost, optim):
+        import ipykernel
+
+        print(ipykernel.__version__)
```
BradyPlanden, line 128: Perhaps an assert instead of a print will ensure this doesn't get missed? 
```suggestion
        assert ipykernel.__version__ >= 0.6
```

BradyPlanden (approved):
Excellent – thanks @NicolaCourtier! Glad to get this one finished, I've added a few suggestions for you to consider. Happy for this to be merged once you are happy. 

As discussed, it would be good to have a normal merge on this to distiguish between the three PR's for debugging in the future.

pybop/plotting/plot2d.py
NicolaCourtier: Let's update this when we allow for more general inputs in the future.

PR #180, opened 2024-02-04
I12b unscented kalman edits

NicolaCourtier:
Edits to the Observer class and Kalman filter implementation.


pybop/observers/observer.py
```diff
@@ -113,3 +132,41 @@ def get_current_time(self) -> float:
...
+        if isinstance(x[0], Parameter):
+            for param in x:
+                inputs[param.name] = param.value
+        else:  # x is an array of parameter values
+            for i, param in enumerate(self.parameters):
+                inputs[param.name] = x[i]
+        self.reset(inputs)
+
+        output = []
+        if hasattr(self, "_dataset"):
```
martinjrobins, line 160: ideally all attributes of a class should be defined in the init. If some attributes are optional then they should either be set to something, or `None`. Then you can test for `None`. Otherwise type checking becomes a pain. This is probably an issue with `Problem` however, so happy to leave it for later.

pybop/observers/unscented_kalman.py
```diff
@@ -27,20 +26,65 @@ class UnscentedKalmanFilterObserver(Observer):
...
+        init_soc=None,
+        x0=None,
     ) -> None:
-        super().__init__(model, inputs, signal)
+        super().__init__(parameters, model, check_model, signal, init_soc, x0)
+        if dataset is not None:
+            self._dataset = dataset.data
+
+            # Check that the dataset contains time and current
+            for name in ["Time [s]", "Current function [A]"] + self.signal:
```
martinjrobins, line 61: why does the dataset need "Current function [A]", the model might not include it or need it (e.g. exponential decay). Also this seems like something that should be defined on the dataset class (e.g. `self._dataset.check_valid()`)

martinjrobins (approved):
Thank you @NicolaCourtier, this looks great :) I have a few suggestion below, they might be more structural things, so happy if we discuss / sort them out later, so I'll hit approve

pybop/observers/observer.py
NicolaCourtier: Agree- this can be sorted after merge! The decision is whether we want to allow the UKF to be run without a dataset (e.g. for testing?) or make it compulsory as it will usually be required.

pybop/observers/unscented_kalman.py
NicolaCourtier: Great minds... I've moved the dataset checks to the Dataset class on another branch (PR #178) so here I just copied what was previously in the Problem class. The current function requirement still needs to be relaxed though.

pybop/observers/observer.py
martinjrobins: I know I did make that optional, but it is a bit weird to call `observe` and not observe anything. If you think it should be compulsory I'm happy to go with that.

PR #183, opened 2024-02-07
Add intermediate model classes

NicolaCourtier:
As discussed, here is an implementation of intermediate model classes to allow the additional model function required for design optiimisation to be included as part of an electrochemical base model - see what you think!

Regarding the nominal capacity estimate, I think it is reliable - I am happy for this to be included as a base echem function called approximate_capacity in combination with the user warnings I've added to the GravimetricEnergyDensity cost.

### DIFF
--- a/examples/notebooks/spm_electrode_design.ipynb
+++ b/examples/notebooks/spm_electrode_design.ipynb
@@ -13340,7 +13340,7 @@
     }
    ],
    "source": [
-    "cost.nominal_capacity(x, cost.problem._model)\n",
+    "cost.problem._model.approximate_capacity(x)\n",
     "pybop.quick_plot(x, cost, title=\"Optimised Comparison\");"
    ]
   },
--- a/pybop/models/empirical/ecm_base.py
+++ b/pybop/models/empirical/ecm_base.py
@@ -1,4 +1,3 @@
-import pybamm
 from ..base_model import BaseModel
 
 
--- a/pybop/models/empirical/ecm_base.py
+++ b/pybop/models/empirical/ecm_base.py
@@ -1,0 +1,30 @@
+import pybamm
+from ..base_model import BaseModel
+
+
+class ECircuitModel(BaseModel):
+    """
+    An intermediate class to extend the functionality of the electrochemical PyBaMM models.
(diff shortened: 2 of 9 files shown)


BradyPlanden:
Hi Nicola,

Thanks for adding this, I think we should have an issue open to track this change moving forward. Can you open one for this PR?

NicolaCourtier:
Hi Brady, I have created issue #186 to clarify the motivation for this PR - let me know if you have any further questions!

examples/notebooks/spm_electrode_design.ipynb
```diff
@@ -13340,7 +13340,7 @@
     }
    ],
    "source": [
-    "cost.nominal_capacity(x, cost.problem._model)\n",
+    "cost.problem._model.approximate_capacity(x)\n",
```
BradyPlanden, line 13343: Future merge reminde: Update this using the model getter in #169

pybop/models/empirical/ecm_base.py
```diff
@@ -0,0 +1,29 @@
+from ..base_model import BaseModel
+
+
+class ECircuitModel(BaseModel):
+    """
+    An intermediate class to extend the functionality of the electrochemical PyBaMM models.
```
BradyPlanden, line 6:
```suggestion
    Overwrites and extends `BaseModel` class for circuit-based models.
```

pybop/models/lithium_ion/echem_base.py
```diff
@@ -0,0 +1,196 @@
+import warnings
+from ..base_model import BaseModel
+
+
+class EChemBaseModel(BaseModel):
+    """
+    An intermediate class to extend the functionality of the electrochemical PyBaMM models.
```
BradyPlanden, line 7:
```suggestion
    Overwrites and extends `BaseModel` class for electrochemical PyBaMM models.
```

BradyPlanden (approved):
Looks great @NicolaCourtier, thanks! 

I've added a comment that I'm hoping will link and remind us to update the private reference to model in the problem class while merging #169. Otherwise just docstring updates.

PR #191, opened 2024-02-09
Add VolumetricEnergyDensity example

NicolaCourtier:
- add VolumetricEnergyDensity
- add cell_volume function
- extend tests
- update Changelog


BradyPlanden (approved):
One spot where we should probably update the tests to match the intended call structure. Otherwise, nice one -  looks great!

PR #197, opened 2024-02-13
Fixes failing observer tests

BradyPlanden:
Adds:

- Longer simulation signal
- Initial SOC = 1.0
- Flips location for higher cost to reduce likelihood of simulation hitting an end condition


martinjrobins (approved):
looks great, thanks @BradyPlanden

Issue #177, opened 2024-01-26
Increase plotting capabilities


NicolaCourtier:
### Feature description

Add a default plot for any number of trajectories and separate the layout properties in StandardPlot from the data.

PR #198, opened 2024-02-16
Additions to #177

BradyPlanden:
This adds the following:

- update subplot trace width via additional `DEFAULT_SUBPLOT_TRACE_OPTIONS` variable
- dataset access method, 
- remove square brackets from plots as per SI standard, 
- updates quick plot for varying length signals, 
- adds `pybop.is_numeric` function for checking input values
- restores design cost exception for optimisation

Encompasses #224. This PR is also a step towards fixing our current failing scheduled tests.


examples/standalone/problem.py
```diff
@@ -14,10 +14,13 @@ def __init__(
         model=None,
         check_model=True,
         signal=None,
+        default_variables=None,
```
NicolaCourtier, line 17: `default-variables` -> `additional_variables`

NicolaCourtier (approved):
Looks ready to merge once the coverage is passing, thanks very much @BradyPlanden!

PR #203, opened 2024-02-16
Add PEP517/518 `pyproject.toml` support

agriyakhetarpal:
Some of the key changes in this PR are as follows:

1. Declarative metadata via TOML tables
2. The `ruff` and `pytest` configurations are included in the newly-added `pyproject.toml`
3. `distutils` is deprecated, so I removed the `setup.py` shim file altogether considering that PyBOP in itself is a pure-Python package
4. Marked `pybop.version` as a private sub-module for `pybop` – this ensures that no one imports `pybop.version.__version__` since the correct and intended way is through accessing `pybop.__version__`, which is unchanged except for the fact that the version is now defined in `pyproject.toml`; the good thing about this is that the package will now not have to be imported via `import pybop` to parse the version string

P.S. I should perhaps look at doing point 4 for PyBaMM too.


pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+]
+maintainers = [
+  {name = "The PyBOP Team"},
+]
+description = "Python Battery Optimisation and Parameterisation"
+readme = "README.md"
+license = { file = "LICENSE" }
+# https://pypi.org/classifiers/
+classifiers = []
+requires-python = ">=3.8, <=3.12"
```
agriyakhetarpal, line 19: I am curious about the Python version bounds, because I copied this verbatim from `setup.py` – since PyBOP does not support Python 3.12 yet, this should have been `<3.12` rather than `<=3.12`. This makes it available, say, for Python 3.12.0, but not Python 3.12.1, so it's not a big deal altogether.

Please let me know if we should explicitly mark it as `<3.12` for now and add 3.12 support in the next release – though I think PyBaMM's own Python version constraint should ensure that functionality for PyBOP is as intended.

pybop/_version.py
agriyakhetarpal, line 1: Marked it as a private module for the reasons stated above, happy to revert the change if needed!

pyproject.toml
BradyPlanden: Nice catch @agriyakhetarpal. Let's update this constraint to be `<3.12` and I'll open an issue to add support for `3.12`.

pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+    "sphinx>=6",
+    "pydata-sphinx-theme",
+    "sphinx-autobuild",
+    "sphinx-autoapi",
+    "sphinx_copybutton",
+    "sphinx_favicon",
+    "sphinx_design",
+    "myst-parser",
+]
+all = ["pybop[plot]"]
```
BradyPlanden, line 40:
```suggestion
all = ["pybop[plot, docs]"]
```

Looks like we missed adding the `docs` optional installation to `all`, let's update it here

BradyPlanden (approved):
Excellent, thanks for the contribution @agriyakhetarpal! Just one spot that's worth updating.

pyproject.toml
agriyakhetarpal: Actually, we want to keep things like `[docs]` and `[dev]` or `[test]` dependencies separate from the `[all]` set. (PyBOP doesn't currently have a `[test]` extra, should I create one and add `pytest`, `pytest-mock`, and others there?)

I am unable to recall where this was referenced in the Scientific Python development guides, but from a rule-of-thumb this recommendation comes from the fact that more optional dependencies included all at once can potentially cause resolution conflicts with the core ones, plus `[docs]` dependencies increase the size of the dependency tree and subsequent interactions with `pip` will not be speedier (though the new tool `uv` counteracts all the latter by just being crazy fast). These extras are not used for the core package as a part of the API either, so they should be placed in isolated environments while testing (including them will break the isolation).

BradyPlanden: Good to know, I wasn't aware of that - let's keep it seperate then!

Taking a second look at the PyBaMM optional dependancies, I think implementing a similar `dev` extra that includes `pytest`, `pytest-mock`, `nox`, `pre-commit` and any others I missed would be a great way forward. 

Aside: I've also been looking at `uv`, adding an optional installation method in the `README` with it should be on our roadmap.

agriyakhetarpal: Thanks for your input! I'll make these changes soon.

`uv` is great with the experiments I've tried locally, we've opened an "up-for-grabs" issue for it over at PyBaMM. See https://github.com/pybamm-team/PyBaMM/issues/3825

agriyakhetarpal: Addressed in https://github.com/pybop-team/PyBOP/pull/203/commits/a6e32a16bf5f62c5f98c0909a872ca1fe395bbb4.

PR #212, opened 2024-02-22
139b Add unit tests for plotting and Thevenin model

NicolaCourtier:


tests/unit/test_models.py
```diff
@@ -11,11 +11,12 @@ class TestModels:
     A class to test the models.
     """
 
+    @pytest.mark.parametrize(
+        "model",
+        [pybop.lithium_ion.SPM(), pybop.lithium_ion.SPMe(), pybop.empirical.Thevenin()],
+    )
```
BradyPlanden, line 17: I think we should be able to use a fixture here to reduce the duplicate parameterisation blocks below.

```suggestion
    @pytest.fixture(params=[
        pybop.lithium_ion.SPM(),
        pybop.lithium_ion.SPMe(),
        pybop.empirical.Thevenin()
    ])
def model(request):
    return request.param
```

tests/unit/test_models.py
```diff
@@ -26,37 +27,49 @@ def test_simulate_without_build_model(self):
...
-        inputs = {
-            "Negative electrode active material volume fraction": 0.52,
-            "Positive electrode active material volume fraction": 0.63,
-        }
+        if isinstance(model, (pybop.lithium_ion.SPM, pybop.lithium_ion.SPMe)):
+            inputs = {
+                "Negative electrode active material volume fraction": 0.52,
+                "Positive electrode active material volume fraction": 0.63,
+            }
+            signal = "Terminal voltage [V]"
```
BradyPlanden, line 54: Since an early release of PyBaMM last year, "Voltage [V]" == "Terminal Voltage [V]". To remain consistent we can move towards using "Voltage [V]" for all signal items.

```suggestion
            signal = "Voltage [V]"
```

tests/unit/test_plots.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @pytest.fixture
+    def model(self):
+        # Define an example model
+        return pybop.lithium_ion.SPM()
+
+    @pytest.mark.unit
+    def test_model_plots(self):
+        # Test plotting of Model objects
+        pass
```
BradyPlanden, line 19: Are we missing test logic here?

tests/unit/test_plots.py
```diff
@@ -0,0 +1,87 @@
...
+                "Current function [A]": values["Current [A]"].data,
+                "Voltage [V]": values["Voltage [V]"].data,
+            }
+        )
+
+        # Generate problem
+        return pybop.FittingProblem(model, parameters, dataset)
+
+    @pytest.mark.unit
+    def test_problem_plots(self):
```
BradyPlanden, line 54: And here?

BradyPlanden (approved):
Thanks for the fast PR @NicolaCourtier! I've added a few comments, but looks great overall. I think we can use a fixture in `test_models.py` to avoid duplicating the parameterize block.

tests/unit/test_plots.py
NicolaCourtier: Also for #177

Issue #210, opened 2024-02-22
Add likelihood classes


BradyPlanden:
### Feature description

As we move to support probablistic methods, a likelihood class that interacts like the cost class is required.

### Possible implementation

The following is a good start towards closing this issue:

- [ ] Base Likelihood Class
- [ ] Gaussian Log Likelihood
- [ ] Gaussian Log Likelihood with know sigma

PR #213, opened 2024-02-22
Update to allow no bounds

NicolaCourtier:
- Use the spm_adam example to test optimisation without bounds, since Adam does not use them
- Allow bounds=None in place of lower and upper bounds
- Add checks to test_parameters
- Raise an error in plot_cost2d if bounds not available


BradyPlanden (commented):
Looks good, thanks for the addition @NicolaCourtier!

Just to confirm, have you checked the performance of the non-gradient based optimisers without bounds added? Bounds are optional from the Pints side, but it would be good to check that they don't affect our solutions.

It would also be good to remove the bounds from the IRPropMin and GradientDescent examples.

BradyPlanden:
Thanks - looks great!

I would consider updating the `test_parameterisation/test_spm_optimisers` with one parameter that doesn't have bounds. That would give us continual check to ensure the lack of bounds on a parameter doesn't effect any current or future optimiser performance.

BradyPlanden (approved):

NicolaCourtier:
Hi @BradyPlanden, this might be worth a second review. I've made the following changes in order to implement your suggestion to have one parameter without bounds in the parameterisation tests:

- add `mean` and `sigma` as properties to the non-Gaussian priors
- pass sigma from the prior through the base problem, cost and optimisation classes to become sigma0 for PINTS
- implement any missing bounds (if some bounds are given) as +/- np.inf
- move the setting of bounds to `__init__` for the SciPy optimisers to match PINTS
- update the bounds check for PINTS PSO and SciPy Differential Evolution which require bounds which are all finite or None
- add `set_bounds` to `Parameter` to allow for consistent updating of bounds, upper and lower (in future, it may be good to pass bounds, sigma and x0 all as part of the parameters object - I think this is related to #210)
- add a default cost value to the `min()` in `plot_convergence` in case of unsuccessful iterations
- update` test_optimisation` and `test_parameterisations` to include tests without bounds
- add property checks to test_priors
- update the Changelog

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
-            lower=[param.bounds[0] for param in self.parameters],
-            upper=[param.bounds[1] for param in self.parameters],
-        )
+        # Set bounds (for all or no parameters)
+        self.bounds = {"lower": [], "upper": []}
+        count = 0
+        for param in self.parameters:
+            if param.bounds is not None:
+                self.bounds["lower"].append(param.bounds[0])
+                self.bounds["upper"].append(param.bounds[1])
```
BradyPlanden, line 54:
```suggestion
                self.bounds["upper"].append(param.bounds[1])
                all_unbounded = False
```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
         self._target = None
 
-        # Set bounds
-        self.bounds = dict(
-            lower=[param.bounds[0] for param in self.parameters],
-            upper=[param.bounds[1] for param in self.parameters],
-        )
+        # Set bounds (for all or no parameters)
+        self.bounds = {"lower": [], "upper": []}
+        count = 0
```
BradyPlanden, line 50:
```suggestion

```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
+        for param in self.parameters:
+            if param.bounds is not None:
+                self.bounds["lower"].append(param.bounds[0])
+                self.bounds["upper"].append(param.bounds[1])
+            else:
+                self.bounds["lower"].append(-np.inf)
+                self.bounds["upper"].append(np.inf)
+                count += 1
+        if count == len(self.parameters):
```
BradyPlanden, line 59:
```suggestion
        if not all_unbounded:
```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
+            self.bounds = None
+
+        # Set initial standard deviation (for all or no parameters)
+        count = 0
+        self.sigma0 = []
+        for param in self.parameters:
+            if hasattr(param.prior, "sigma"):
+                self.sigma0.append(param.prior.sigma)
```
BradyPlanden, line 67:
```suggestion
                self.sigma0.append(param.prior.sigma)
                has_sigma = True
```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
+                self.bounds["upper"].append(param.bounds[1])
+            else:
+                self.bounds["lower"].append(-np.inf)
+                self.bounds["upper"].append(np.inf)
+                count += 1
+        if count == len(self.parameters):
+            self.bounds = None
+
+        # Set initial standard deviation (for all or no parameters)
+        count = 0
```
BradyPlanden, line 63:
```suggestion

```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
+        self.sigma0 = []
+        for param in self.parameters:
+            if hasattr(param.prior, "sigma"):
+                self.sigma0.append(param.prior.sigma)
+            else:
+                count += 1
```
BradyPlanden, line 69:
```suggestion

```

pybop/_problem.py
```diff
@@ -45,11 +45,30 @@ def __init__(
...
+        if count > 0:
```
BradyPlanden, line 70:
```suggestion
        if not has_sigma:
```

pybop/optimisers/pints_optimisers.py
```diff
@@ -116,12 +117,16 @@ class PSO(pints.PSO):
     """
 
     def __init__(self, x0, sigma0=0.1, bounds=None):
-        if bounds is not None:
+        if bounds is None:
+            self.boundaries = None
+        elif not np.all(np.isfinite(sum(bounds.values(), []))):
```
BradyPlanden, line 122: I think this might do what you are wanting in a more pythonic way.
```suggestion
        elif not all(np.isfinite(value) for sublist in bounds.values() for value in sublist):
```

pybop/optimisers/scipy_optimisers.py
```diff
@@ -137,12 +136,21 @@ class SciPyDifferentialEvolution(BaseOptimiser):
...
         super().__init__()
-        self.bounds = bounds
         self.strategy = strategy
         self._max_iterations = maxiter
         self._population_size = popsize
 
-    def _runoptimise(self, cost_function, x0=None, bounds=None):
+        if bounds is None:
+            raise ValueError("Bounds must be specified for differential_evolution.")
+        elif not np.all(np.isfinite(sum(bounds.values(), []))):
```
BradyPlanden, line 145: Like above:
```suggestion
        elif not all(np.isfinite(value) for sublist in bounds.values() for value in sublist):
```

pybop/parameters/parameter.py
```diff
@@ -120,3 +118,27 @@ def set_margin(self, margin):
...
+
+        Raises
+        ------
+        ValueError
+            If the lower bound is not strictly less than the upper bound, or if
+            the margin is set outside the interval (0, 1).
+        """
+        if bounds is not None:
+            if bounds[0] >= bounds[1]:
+                raise ValueError("Lower bound must be less than upper bound")
```
BradyPlanden, line 140:
```suggestion
        if bounds is not None and bounds[0] >= bounds[1]:
```

pybop/parameters/parameter.py
```diff
@@ -120,3 +118,27 @@ def set_margin(self, margin):
...
+            self.lower_bound = bounds[0]
+            self.upper_bound = bounds[1]
+
```
BradyPlanden, line 143:
```suggestion
            self.upper_bound = bounds[1]
        else:
            raise ValueError("Lower bound must be less than upper bound")

```

pybop/plotting/plot_convergence.py
```diff
@@ -29,7 +30,7 @@ def plot_convergence(
 
     # Compute the minimum cost for each iteration
     min_cost_per_iteration = [
-        min(cost_function(solution) for solution in log_entry)
+        min((cost_function(solution) for solution in log_entry), default=np.inf)
```
BradyPlanden, line 33: We should just save the cost value during optimisation and plot it here. At the moment, this plot takes quite a while since it completes the optimisation a second time. 

Does the `solution` object contain the cost value? We can push this to a future PR, but should probably open an issue if we don't fix it here.

tests/integration/test_parameterisations.py
```diff
@@ -147,6 +157,16 @@ def spm_two_signal_cost(self, parameters, model, x0):
     )
     @pytest.mark.integration
     def test_multiple_signals(self, optimiser, spm_two_signal_cost, x0):
+        # Some optimisers require a complete set of bounds
+        if optimiser in [pybop.SciPyDifferentialEvolution]:
```
BradyPlanden, line 161: Does the below represent what we expect end-users to do? It seems like it could be pushed into the codebase and abstracted from the user if so.

BradyPlanden (commented):
Thanks for adding in the additional code for unbounded optimisation @NicolaCourtier! I have a few comments in areas that I think we can improve code comprehension. All in all it's looking very good!

tests/integration/test_parameterisations.py
NicolaCourtier: No, this should be updated when we move to using a Parameters object instead of a list of parameters! Here, it is written this way in order to re-use the cost fixture.

pybop/plotting/plot_convergence.py
NicolaCourtier: This edit is simply to cope with failures in the evaluation of the cost function which, in unusual cases, left an empty array with no minimum value.

NicolaCourtier: I've copied your comment to the plotting PR: https://github.com/pybop-team/PyBOP/pull/178/files#r1507377660

PR #215, opened 2024-02-23
Update release workflow

BradyPlanden:
This adds:

- A [release_workflow.md](https://github.com/pybop-team/PyBOP/blob/06ac3743c3a95fd3476920beb35a73499316aace/.github/release_workflow.md) to clarify the release steps and workflow
- Updates the release_action.yaml for trusted publishing.

### DIFF
--- a/.github/release_workflow.md
+++ b/.github/release_workflow.md
@@ -1,0 +1,34 @@
+# Release Workflow
+
+This document outlines the release workflow for publishing to PyPI and TestPyPI using GitHub Actions.
+
+## Creating a New Release
+
+To create a new release, follow these steps:
+
+1. **Prepare the Release:**
+   - Create a new branch for the release (i.e. `v24.XX`) from `develop`.
+   - Increment the following:
+         - The version number in the `pyproject.toml` file following CalVer versioning.
+         - The`CHANGELOG.md` version with the changes for the new version.
+   - Open a PR to the `main` branch. Once the PR is merged, proceed to the next step.
+
+2. **Tag the Release:**
+   - Create a new Git tag for the release. For a full release, use a tag like `v24.2`. For a release candidate, use a tag like `v24.2-rc.1`.
+   - Push the tag to the remote repository: `git push origin <tag_name>`.
+
+3. **Create a GitHub Release:**
+   - Go to the "Releases" section of on GitHub.
+   - Click "Draft a new release."
+   - Enter the tag you created in the "Tag version" field.
+   - Fill in the release title and description. Add any major changes and link to the `CHANGELOG.md` for a list of total changes.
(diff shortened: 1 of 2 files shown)


.github/release_workflow.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. **Prepare the Release:**
+   - Create a new branch for the release (i.e. `v24.XX`) from `develop`.
+   - Increment the following:
+         - The version number in the `pyproject.toml` file following CalVer versioning.
+         - The`CHANGELOG.md` version with the changes for the new version.
+   - Open a PR to the `main` branch. Once the PR is merged, proceed to the next step.
+
+2. **Tag the Release:**
+   - Create a new Git tag for the release. For a full release, use a tag like `v24.2`. For a release candidate, use a tag like `v24.2-rc.1`.
```
agriyakhetarpal, line 17: I would suggest removing the hyphen (`-`) between the string and the numeric characters, in lieu of PEP 440 versioning compliance, available here: https://peps.python.org/pep-0440/#pre-releases

```suggestion
   - Create a new Git tag for the release. For a full release, use a tag like `v24.2`. For a release candidate, use a tag like `v24.2rc.1`.
```

.github/workflows/release_action.yaml
```diff
@@ -53,6 +52,9 @@ jobs:
         path: dist/
     - name: Publish distribution 📦 to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
agriyakhetarpal, line 57: I don't think these are required, since the `pypi` environment is already configured for this job (the use of a short-lived API token eliminates the need for API tokens permanently configured in the secrets).

agriyakhetarpal (approved):
Thanks @BradyPlanden, I have read through this and can approve with two comments. I would suggest adding a condition to stop the `publish` job from being run on forks, for security reasons, just in case.

.github/workflows/release_action.yaml
BradyPlanden: Ah fair, I was following the PyBaMM release workflow and thought the addition would help with verification. Taking a second look at the [marketplace](https://github.com/marketplace/actions/pypi-publish#specifying-a-different-username) PyPi release action, they don't seem to be recommended anymore. Perhaps it should be removed from PyBaMM as well.

agriyakhetarpal: Ah yes, we intend to remove those soon – it hasn't had a pair of eyes looking at it yet: https://github.com/pybamm-team/PyBaMM/pull/3358

PR #217, opened 2024-02-26
Add PyBaMM IDAKLU solver, compiled from source

agriyakhetarpal:
Till the time PyBaMM puts out a new release to fix the IDAKLU solver in its PyPI wheels, this PR shall help compile it from source in order for it to be used by PyBOP.

I have merged the latest changes from `develop` in this branch since #176 had not been updated in a while. This PR shall update the branch there upon getting merged, since this doesn't target the `develop` branch.


BradyPlanden (commented):
Thanks for opening this PR @agriyakhetarpal. I've approved the workflow, it looks like we are running into a few issues builting from source.

agriyakhetarpal:
Actually, building from source is working on the coverage tests, but it's trying to assume `pybind11` is installed and then trying to run tests from its top-level directory and for PyBaMM. We have wanted to remove the git-clone for pybind11 in https://github.com/pybamm-team/PyBaMM/pull/3560 but it has been stalled for many reasons.

I have pushed d63fc769bca0a35666514e4c306a463278a8674f which should fix this.

agriyakhetarpal:
@BradyPlanden this should be ready for review now. Windows wheels for PyBaMM take ~20 minutes to build, and this can be reduced to ~4 minutes with caching – this is not worth using since they are broken as previously discussed. The Linux and macOS PyBaMM wheels are being compiled in ~3 minutes and work well – all of the test failures are coming from the PyBOP suite and you may take this up to fix up the test cases, either here, or merge this PR and then continue on #176, which is where these commits shall show up and then can be merged into the `develop` branch when ready.

BradyPlanden (approved):
Thanks for taking the time to sort this @agriyakhetarpal. All looks good to me! I'll sort out the PyBOP failing tests in #176.

PR #224, opened 2024-03-01
Datatype restructure for optimisation objects

BradyPlanden:
I'm raising this commit to a PR as I believe there are enough changes to the backend to warrant a review at this stage. This started as an update to the plotting class to include alternative x-axis variables, but moved towards refactoring the pybop backend to make this possible.

This PR restructures PyBOP's datatypes for model output (`y` & `dy`). As such, knock-on changes to the problem and cost classes were required. The main benefits of these changes are to move away from numpy array objects where possible and replace them with dictionaries. This PR achieves the following:

- Order independent access from the dictionary. This benefit can be seen in the updates to the plotting classes.
- Clean multi-signal integration, as iteration over the `problem.signal` becomes the only requirement.
- A reduction in the datatype castings we needed within `model.simulate` and the corresponding cost function classes.
- Adds a `problem.additional_variables` object to capture variables needed within the design problem class, as well as plotting.
- Updates tests for the new datatypes
- Fixes the integration tests as the previous assertions where compared to `x0` and not the groundtruth. They have been updated to sample from a random normal distribution to set the groundtruth, with `x0` also randomly sampled too. This now represents the true optimiser performance.

@NicolaCourtier @martinjrobins, there are quite a few changes in this one, and I would appreciate it if you both could take a good look at any potential bugs or improvements. I've spent a lot of time double-checking the cost function calculations, but it's very possible that I've missed something.

At the moment, the `UKF` tests are failing, I need to take a second look at them. @martinjrobins, if you can take a look as well that would be appreciated.

### DIFF
--- a/pybop/_problem.py
+++ b/pybop/_problem.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pybop
 
 
 class BaseProblem:
@@ -27,6 +28,7 @@ class BaseProblem:
         model=None,
         check_model=True,
         signal=["Voltage [V]"],
+        default_variables=[],
         init_soc=None,
         x0=None,
     ):
@@ -44,6 +46,11 @@ class BaseProblem:
         self.n_outputs = len(self.signal)
         self._time_data = None
         self._target = None
+
+        if isinstance(model, (pybop.BaseModel, pybop.lithium_ion.EChemBaseModel)):
+            self.default_variables = default_variables
+        else:
+            self.default_variables = []
 
         # Set bounds
         self.bounds = dict(
@@ -148,10 +155,13 @@ class FittingProblem(BaseProblem):
         dataset,
         check_model=True,
         signal=["Voltage [V]"],
+        default_variables=["Time [s]", "Discharge capacity [A.h]"],
         init_soc=None,
         x0=None,
     ):
-        super().__init__(parameters, model, check_model, signal, init_soc, x0)
+        super().__init__(
+            parameters, model, check_model, signal, default_variables, init_soc, x0
+        )
         self._dataset = dataset.data
         self.x = self.x0
 
@@ -161,12 +171,13 @@ class FittingProblem(BaseProblem):
         # Unpack time and target data
         self._time_data = self._dataset["Time [s]"]
         self.n_time_data = len(self._time_data)
-        target = [self._dataset[signal] for signal in self.signal]
-        self._target = np.vstack(target).T
+        self._target = {signal: self._dataset[signal] for signal in self.signal}
 
         # Add useful parameters to model
         if model is not None:
             self._model.signal = self.signal
+            self._model.default_variables = self.default_variables
+            self._model.n_parameters = self.n_parameters
             self._model.n_outputs = self.n_outputs
             self._model.n_time_data = self.n_time_data
 
@@ -193,14 +204,14 @@ class FittingProblem(BaseProblem):
         y : np.ndarray
             The model output y(t) simulated with inputs x.
         """
-        if (x != self.x).any() and self._model.matched_parameters:
+        if np.any(x != self.x) and self._model.matched_parameters:
             for i, param in enumerate(self.parameters):
                 param.update(value=x[i])
 
             self._model.rebuild(parameters=self.parameters)
             self.x = x
 
-        y = np.asarray(self._model.simulate(inputs=x, t_eval=self._time_data))
+        y = self._model.simulate(inputs=x, t_eval=self._time_data)
 
         return y
 
@@ -229,7 +240,7 @@ class FittingProblem(BaseProblem):
             t_eval=self._time_data,
         )
 
-        return (np.asarray(y), np.asarray(dy))
+        return (y, np.asarray(dy))
 
 
 class DesignProblem(BaseProblem):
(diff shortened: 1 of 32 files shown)


pybop/_problem.py
```diff
@@ -45,6 +47,11 @@ def __init__(
         self._time_data = None
         self._target = None
 
+        if isinstance(model, (pybop.BaseModel, pybop.lithium_ion.EChemBaseModel)):
```
martinjrobins, line 50: this is unexpected, why is `default_variables` only used for these particular models?

martinjrobins: I don't really understand what a `default_variables` is either! But perhaps a docstring would fix that :)

martinjrobins (approved):
Hi @BradyPlanden, I've had a quick look through and all seems good, happy with the change to dictionaries naming the `y` and `dy` outputs. I wasn't sure what `default_variables` was, but I don't really know how the design stuff works so that is probably why

pybop/_problem.py
BradyPlanden: Thanks @martinjrobins. Good shout, I'll add a docstring. `default_variables` can be viewed as the variables needed for plotting and the inner working of the problem + cost classes. For example, we have had `Time [s]` sprinkled around the classes, inaddition to passing it from the `signal` variable occassionally. `default_variables` captures these variables without needing them added to `signal`. 

The logic can probably improve in this situation, it's currently protecting against calling pybamm specific variables (`'Time [s]'`, etc.) when using a standalone or alternative model that doesn't have those keys.

tests/unit/test_standalone.py
```diff
@@ -53,13 +53,18 @@ def test_standalone_problem(self):
 
         # Test the Problem with a Cost
         rmse_cost = pybop.RootMeanSquaredError(problem)
-        x = rmse_cost([1, 2])
+        rmse_x = rmse_cost([1, 2])
+        rmse_grad_x = rmse_cost.evaluateS1([1, 2])
 
-        np.testing.assert_allclose(x, 3.138, atol=1e-2)
+        np.testing.assert_allclose(rmse_x, 3.05615, atol=1e-2)
```
BradyPlanden, line 59: This change was a bit worrisome, since the value should remain constant. I did a comparison with Pints' logic and it returned the updated value as well, pointing towards a bug in the previous RMSE calculation.

pybop/_problem.py
BradyPlanden: This has been updated to `additional_variables` with a docstring added.

NicolaCourtier (commented):
I think this PR is heading in a very useful direction! Thanks @BradyPlanden. Here are some initial comments.

1. On the naming of `additional_variables` - can you make it clear what these variables are in addition to, in the description? Are they optional extras for plotting? Or perhaps `required_variables`, i.e. the variables that are required for the specific optimisation problem?

2. Do you think that `Discharge capacity [A.h]` will be required for all types of design problem? I would expect different requirements for different design costs. Could further variables be added as part of the Cost?

3. I'm also not sure why the passing of `additional_variables` is limited to the Base and EChem models, why not the PyBaMM ECM models for example? If this logic is required, please add a warning for times when the input is not being used.

4. In the cost functions, it is not clear to me why these two variables have been selected:
```
if key not in ["Time [s]", "Discharge capacity [A.h]"]:
```
and
```
if signal not in ["Time [s]", "Discharge capacity [A.h]"]
```
The original check was designed to cope with simulations that terminate early (which we expect in several situations such as reaching a voltage limit during the simulation). However, as written I think it is too general and could hide other, unexpected errors.

5. Please can you increase the coverage and check why one of the tests appears to be failing?

pybop/costs/fitting_costs.py
```diff
@@ -39,10 +39,23 @@ def _evaluate(self, x, grad=None):
         """
         prediction = self.problem.evaluate(x)
 
-        if len(prediction) < len(self._target):
-            return np.float64(np.inf)  # simulation stopped early
+        for key in prediction:
+            if key not in ["Time [s]", "Discharge capacity [A.h]"]:
```
BradyPlanden, line 43: Filtering at this point is to avoid looping across the whole prediction dictionary, on second viewing it might be better to filter based on the `additional_variable` object.

Although, it might be worth looping across the whole dictionary in some cases specifically if ithe problem is standalone and hasn't been verified.

pybop/costs/fitting_costs.py
```diff
@@ -39,10 +39,23 @@ def _evaluate(self, x, grad=None):
...
+                if len(prediction.get(key, [])) != len(self._target.get(key, [])):
+                    return np.float64(np.inf)  # prediction doesn't match target
+
+        e = np.array(
+            [
+                np.sqrt(np.mean((prediction[signal] - self._target[signal]) ** 2))
+                for signal in prediction
+                if signal not in ["Time [s]", "Discharge capacity [A.h]"]
```
BradyPlanden, line 51: We don't need the additional variables in error computation. Likewise to above, filtering based on the additional variables and `Time [s]` might be a clearer choice.

pybop/costs/fitting_costs.py
NicolaCourtier: The relevant variables here are those in the signal. For functioning models, it should only be necessary to check the length of one output (as all outputs should have the same length), so I suggest:

```
if len(prediction.get(signal[1], [])) < len(self._target.get(signal[1], [])):
     return np.float64(np.inf)  # simulation stopped early
```
It would also be ok to loop over all the signals.

pybop/costs/fitting_costs.py
NicolaCourtier: I think that users should have free choice of variables to include in `signal`. We would have to add extra warnings if we are to disallow some variables.

PR #230, opened 2024-03-08
Make likelihoods a type of cost

NicolaCourtier:
# Description

Suggestion to make the new likelihood classes a type of cost.

## Issue reference
Regarding #210 

## Type of change


pybop/costs/_likelihoods.py
```diff
@@ -1,27 +1,19 @@
...
         self._n_parameters = problem.n_parameters
-        self._target = problem._target
-
-    def __call__(self, x):
-        """
-        Calls the problem.evaluate method and calculates
-        the log-likelihood
-        """
-        raise NotImplementedError
+        self.log_likelihood = problem
```
BradyPlanden, line 16: This can be removed as it was only required for the ProbabilityCost() wrapper.
```suggestion

```

tests/unit/test_likelihoods.py
```diff
@@ -75,12 +75,6 @@ def test_base_likelihood_init(self, problem):
         assert likelihood._n_parameters == 1
         assert np.array_equal(likelihood._target, problem._target)
 
-    @pytest.mark.unit
```
BradyPlanden, line 78: I think we still want this test? Looking at the above class, `BaseLikelihood` still returns NotImplemented from __call__(). It would be good to keep a test on that.

pybop/_optimisation.py
```diff
@@ -40,27 +40,24 @@ class Optimisation:
...
-        x0=None,
         optimiser=None,
         sigma0=None,
         verbose=False,
         physical_viability=True,
         allow_infeasible_solutions=True,
     ):
         self.cost = cost
-        self.x0 = x0
+        self.x0 = cost.x0
```
BradyPlanden, line 50: As mentioned above. This keeps it open :)
```suggestion
        self.x0 = x0 or cost.x0
```

pybop/_optimisation.py
```diff
@@ -40,27 +40,24 @@ class Optimisation:
     def __init__(
         self,
         cost,
```
BradyPlanden, line 42: I think we want users to overwrite the initial conditions, right? This modification should keep that open.

```suggestion
        cost,
        x0=None,
```

BradyPlanden (approved):
Looks great, thanks for the additions @NicolaCourtier! I've added a few comments, I'm happy to update these in #210 though :) Let me know if any thoughts on them!

Issue #233, opened 2024-03-11
Experiment breaks geometric fitting workflow


BradyPlanden:
### Python Version

3.11

### Describe the bug

Generating synthetic data from `pybop.experiment` breaks the fitting for geometric parameters. 

### Steps to reproduce the behaviour

```python
import pybop
import numpy as np

# Define model
parameter_set = pybop.ParameterSet.pybamm("Chen2020")
model = pybop.lithium_ion.SPM(parameter_set=parameter_set)

# Fitting parameters
parameters = [
    pybop.Parameter(
        "Negative particle radius [m]",
        prior=pybop.Gaussian(6e-06, 0.1e-6),
        bounds=[1e-6, 9e-6],
        true_value=parameter_set["Negative particle radius [m]"],
    ),
    pybop.Parameter(
        "Positive particle radius [m]",
        prior=pybop.Gaussian(4.5e-06, 0.1e-6),
        bounds=[1e-6, 9e-6],
        true_value=parameter_set["Positive particle radius [m]"],
    ),
]

# Generate data
sigma = 0.001
init_soc = 1  # start from full charge
experiment = pybop.Experiment(
    [
        "Discharge at 3A until 3.8 V (2 seconds period)",
    ],
)

values = model.predict(experiment=experiment, init_soc=init_soc)
corrupt_values = values["Voltage [V]"].data + np.random.normal(
    0, sigma, len(values["Time [s]"].data)
)
# Form dataset
dataset = pybop.Dataset(
    {
        "Time [s]": values["Time [s]"].data,
        "Current function [A]": values["Current [A]"].data,
        "Voltage [V]": corrupt_values,
    }
)

# Generate problem, cost function, and optimisation class
problem = pybop.FittingProblem(model, parameters, dataset, init_soc=init_soc)
cost = pybop.SumSquaredError(problem)
optim = pybop.Optimisation(cost, optimiser=pybop.CMAES)
optim.set_max_iterations(100)

# Run the optimisation
x, final_cost = optim.run()
print(
    "True parameters:",
    [
        parameters[0].true_value,
        parameters[1].true_value,
    ],
)
print("Estimated parameters:", x)

# Plot the timeseries output
pybop.quick_plot(x, cost, title="Optimised Comparison")

# Plot convergence
pybop.plot_convergence(optim)

# Plot the parameter traces
pybop.plot_parameters(optim)

# Plot the cost landscape
pybop.plot_cost2d(cost, steps=15)

[...]

PR #240, opened 2024-03-14
Fix workflow for changing experiment

NicolaCourtier:
# Description

Fix for issue #233. The current function was not being set during the call to `set_params` unless `unmatched_parameters` were present.

### DIFF
--- a/pybop/models/base_model.py
+++ b/pybop/models/base_model.py
@@ -160,7 +160,7 @@ class BaseModel:
             for i in self.fit_keys:
                 self._parameter_set[i] = "[input]"
 
-        if self.dataset is not None and self.non_matched_parameters:
+        if self.dataset is not None and (not rebuild or self.non_matched_parameters):
             if "Current function [A]" not in self.fit_keys:
                 self._parameter_set["Current function [A]"] = pybamm.Interpolant(
                     self.dataset["Time [s]"],
--- a/pybop/models/base_model.py
+++ b/pybop/models/base_model.py
@@ -92,7 +92,7 @@ class BaseModel:
         self.dataset = dataset
         self.parameters = parameters
         if self.parameters is not None:
-            self.update_classified_parameters(self.parameters)
+            self.classify_and_update_parameters(self.parameters)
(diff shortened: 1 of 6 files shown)


tests/integration/test_changes.py
```diff
@@ -0,0 +1,103 @@
...
+    @pytest.mark.integration
+    def test_changing_experiment(self, parameter):
+        # Change the experiment and check that the results are different.
+
+        parameter_set = pybop.ParameterSet.pybamm("Chen2020")
+        parameters = [parameter]
+        init_soc = 0.5
+        model = pybop.lithium_ion.SPM(parameter_set=parameter_set)
+
+        t_eval = [0, 3600]  # Default 1C discharge to cut-off voltage
```
BradyPlanden, line 39: I believe this will result in the prediction output to be of length 100. I think this stems from the default result pybamm provides if a sampling time is not requested. To ensure the assert below is purely based on element differences, I think we want something like this instead:
```suggestion
        t_eval = np.arange(0, 3600, 2)  # Default 1C discharge to cut-off voltage
```

To ensure the shapes are always the same, it might be a good idea to shrink the simulation time to be a bit below 3600.

tests/integration/test_changes.py
BradyPlanden, line 1: A clearer name could be: `test_model_experiment_changes.py` with a corresponding class name of`TestModelAndExperimentChanges`.

BradyPlanden (approved):
Looks great — thanks for the fast bugfix @NicolaCourtier! A few small suggestions, happy for you to merge when ready.

pybop/models/base_model.py
```diff
@@ -261,9 +258,6 @@ def set_parameter_classification(self, parameters):
             self._unprocessed_parameter_set = self._parameter_set
             self.geometry = self.pybamm_model.default_geometry
 
-        if self.non_matched_parameters:
```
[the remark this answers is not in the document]
NicolaCourtier: Yes, I thought that `fit_keys` should stay as the complete list of fitting parameter names, and we can use `non_matched_parameters.keys()` directly when needed.

PR #243, opened 2024-03-19
Patch: benchmark solution tracking

BradyPlanden:
The benchmarks are currently failing since the object returned from `track_parameterisation` is of type `tuple`. `asv` only supports single element objects, so this change updates the `track` methods to track each element separately.

To minimise the number of times the benchmark is ran, I've split the benchmark into separate timing and tracking classes. This split minimizes the number of times the parameterisation benchmark is ran, unfortunately three times seems to be the minimum (`time_parameterisation`, `track_x1`, and `track_x2`) as `asv` doesn't support vector/list objects.

### DIFF
--- a/benchmarks/benchmark_track_parameterisation.py
+++ b/benchmarks/benchmark_track_parameterisation.py
@@ -3,7 +3,7 @@ import numpy as np
 from .benchmark_utils import set_random_seed
 
 
-class BenchmarkParameterisation:
+class BenchmarkTrackParameterisation:
     param_names = ["model", "parameter_set", "optimiser"]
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
I am not fully sure I follow what is going on, but I think it makes sense. We could split up the benchmarks into two parallel CI jobs too if they are taking too much time right now.

BradyPlanden:
Thanks for the review, for reference I've updated the original comment to be more descriptive of the problem. Unfortunately, I haven't found a way to track the optimisation output of our methods and time them at the same time.

Issue #79, opened 2023-10-31
Add BPX support 


BradyPlanden:
### Feature description

This issue is for adding BPX support for import/export of parameter sets. Initial thoughts on the API:

To export:
```python
bpx = pybop.BPX(results=results, parameter_set=parameter_set)
bpx.save(file='test_params.json')
```

To import:
```python
parameter_set = pybop.BPX.import(file='test_params.json')
```

PR #246, opened 2024-03-19
Add BPX support

NicolaCourtier:
# Description

Add support for importing BPX JSON files as parameter sets.

## Issue reference
Adds feature request #79.

## Type of change
- [x] New Feature: A non-breaking change that adds new functionality.


pybop/parameters/parameter_set.py
```diff
@@ -60,6 +60,36 @@ def import_parameters(self, json_path=None):
...
+
+        Raises
+        ------
+        FileNotFoundError
+            If the specified JSON file cannot be found.
+        """
+
+        # Read JSON file
+        if not self.params and self.json_path:
+            self.params = pybamm.ParameterValues.create_from_bpx(self.json_path)
```
BradyPlanden, line 90: It would be great to raise an error here. For example, something like the below.

```suggestion
        if not self.params and self.json_path:
            self.params = pybamm.ParameterValues.create_from_bpx(self.json_path)
        else:
            raise ValueError("Parameter set already construct, or path to json file not provided.")
```

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
   "numpy>=1.16",
   "scipy>=1.3",
   "pints>=0.5",
+  "bpx>=0.4",
```
BradyPlanden, line 25: We might be able to relax this dependancy, did you happen to test with earlier versions of BPX?

BradyPlanden (commented):
Thanks for the addition @NicolaCourtier, on a whole this looks great! I've made two small comments that are worth investigating before merging. Let me know what you think.

pyproject.toml
NicolaCourtier: I didn't test with earlier versions but since this functionality is taken from PyBaMM, I think it's best to have the same restriction here. PyBaMM has: `"bpx>=0.4.0"`.

tests/unit/test_models.py
```diff
@@ -114,6 +114,15 @@ def test_rebuild(self, model):
                 initial_built_model, attribute
             )
 
+    @pytest.mark.unit
+    def test_parameter_set_definition(self):
+        # Test initilisation with different types of parameter set
+        param_dict = {"Nominal cell capacity [A.h]": 5}
+        pybop.BaseModel(parameter_set=None)
+        pybop.BaseModel(parameter_set=param_dict)
```
BradyPlanden, line 122: It would be a good idea to add asserts after each construction.

BradyPlanden (approved):
Thank @NicolaCourtier.

LGTM, just one suggestion on the unit tests.

PR #252, opened 2024-03-22
Increment `PyBaMM > 23.5`

BradyPlanden:
# Description

Increments PyBaMM > 23.5 to ensure `pybamm.Thevenin` is usable within PyBOP and fixes the current failing scheduled tests.

## Issue reference
Fixes #251 

## Review
Before you mark your PR as ready for review, please ensure that you've considered the following:
- Updated the [CHANGELOG.md](https://github.com/pybop-team/PyBOP/blob/develop/CHANGELOG.md) in reverse chronological order (newest at the top) with a concise description of the changes, including the PR number.
- Noted any breaking changes, including details on how it might impact existing functionality.

## Type of change
- [ ] New Feature: A non-breaking change that adds new functionality.
- [ ] Optimization: A code change that improves performance.
- [X] Bug Fix: A non-breaking change that addresses an issue.
- [ ] Documentation: Updates to documentation or new documentation for new features.
- [ ] Refactoring: Non-functional changes that improve the codebase.
- [ ] Style: Non-functional changes related to code style (formatting, naming, etc).
- [x] Testing: Additional tests to improve coverage or confirm functionality.
- [ ] Other: (Insert description of change)

# Key checklist:

- [X] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybop-team/PyBOP/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [X] All unit tests pass: `$ nox -s tests`
- [X] The documentation builds: `$ nox -s docs`

You can run integration tests, unit tests, and doctests together at once, using `$ nox -s quick`.

## Further checks:
- [ ] Code is well-commented, especially in complex or unclear areas.
- [ ] Added tests that prove my fix is effective or that my feature works.
- [x] Checked that coverage remains or improves, and added tests if necessary to maintain or increase coverage.

Thank you for contributing to our project! Your efforts help us to deliver great software.

PR #285, opened 2024-04-10
Increment `Python > 3.8`

### LINKED ISSUE #284: Drop Python 3.8
### Feature description

As PyBaMM has now [dropped Python 3.8 ](https://github.com/pybamm-team/PyBaMM/pull/3961) ahead of it's end of life support in August, we should follow for the next release.

### Motivation

_No response_

### Possible implementation

_No response_

### Additional context

_No response_

BradyPlanden:
# Description

Increments Python version to be > 3.8.

## Issue reference
Fixes #284 

## Review

## Type of change
- [X] Other: Increment Python version

# Key checklist:

- [X] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybop-team/PyBOP/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [X] All unit tests pass: `$ nox -s tests`
- [X] The documentation builds: `$ nox -s doctest`

## Further checks:

### DIFF
--- a/scripts/ci/build_matrix.sh
+++ b/scripts/ci/build_matrix.sh
@@ -9,7 +9,7 @@
 
 # To update the matrix, the variables below can be modified as needed.
 
-python_version=("3.8" "3.9" "3.10" "3.11" "3.12")
+python_version=("3.9" "3.10" "3.11" "3.12")
 os=("ubuntu-latest" "windows-latest" "macos-latest")
 # This command fetches the last three PyBaMM versions excluding release candidates from PyPI
 pybamm_version=($(curl -s https://pypi.org/pypi/pybamm/json | jq -r '.releases | keys[]' | grep -v rc | tail -n 3 | paste -sd " " -))
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -59,7 +59,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         exclude: # We run the coverage tests on Ubuntu with Python 3.12
           - os: ubuntu-latest
             python-version: "3.12"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+- [#285](https://github.com/pybop-team/PyBOP/pull/285) - Drop support for Python 3.8.
 - [#273](https://github.com/pybop-team/PyBOP/pull/273) - Adds notebooks to nox examples session and updates CI workflows for change.
 - [#250](https://github.com/pybop-team/PyBOP/pull/250) - Adds DFN, MPM, MSMR models and moves multiple construction variables to BaseEChem. Adds exception catch on simulate & simulateS1.
 - [#241](https://github.com/pybop-team/PyBOP/pull/241) - Adds experimental circuit model fitting notebook with LG M50 data.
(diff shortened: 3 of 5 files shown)


scripts/ci/build_matrix.sh
```diff
@@ -9,7 +9,7 @@
 
 # To update the matrix, the variables below can be modified as needed.
 
-python_version=("3.8" "3.9" "3.10" "3.11" "3.12")
+python_version=("3.9" "3.10" "3.11" "3.12")
```
agriyakhetarpal, line 12: This comment is not a review, but more of a question about our policy for dropping these versions. Is the plan to follow PyBaMM strictly and create another release for PyBOP as soon as possible after PyBaMM's own release?

There could be a case where PyBaMM drops support for Python v3.A in its new version v2X.Y, and PyBOP v2M.N might still support Python v3.A, which means one matrix target will start to fail and that will create a signal to drop the version for us here at PyBOP. Users running PyBOP v2M.N at that time on Python v3.A will notice that the new PyBaMM v2X.Y will not be downloaded by `pip` because of the Python version constraint, and therefore they would be using an older version, say PyBaMM v2X.W. However, those on newer versions of Python, say, v3.B, will not face this constraint because it will be supported by both PyBOP and PyBaMM (assuming that both packages will support an array of at least four Python versions at a time), and therefore v3.B will compel `pip` to download PyBaMM v2X.Y + PyBOP v2M.N. Here, there could be breaking changes for PyBaMM which PyBOP would not be supporting, and therefore PyBOP would need to put out a new release after v2M.N with a new identifier v2M.O to support the new PyBaMM v2X.Y along with previous PyBaMM versions (but if it's not possible to support older ones, then drop one and support one at the same time) and drop support for Python v3.A (and add support for newer Python versions supported by PyBaMM). Therefore, I have these comments:

1. Maybe PyBOP should consider upper-pinning PyBaMM in its dependencies (in addition to lower bounds) to avoid such behaviour
2. The answer is also highly dependent on how well PyBaMM documents its breaking changes and features, and also on the extent of the breaking changes (it's unlikely that, say, a PyBaMM solver changes its functionality or anything, but other things like parameter sets with the BPX standardisation could – that I do mot have a lot of information about). It would also be reliant on how users are going to use PyBaMM and PyBOP together and how reliant the latter will be on the former in terms of certain functionality and offering convenient features.
3. I suppose these issues would only be encountered on those that are running _edge_ versions of Python, and not those intermediate ones that will continue to stay in support for yesrs on stretch and will be dropped after a considerable amount of releases (i.e., not in the immediate next releases).

I guess I am just concerned about the synchronicity between PyBaMM and PyBOP; there is a chance that my concerns could be misplaced, but I hope that what I am asking makes a bit of sense and is something to ponder about!

P.S. Adjusting this script to cater to new Python and PyBaMM versions or ignore specific ones is quite trivial, of course.

agriyakhetarpal: P.P.S. the answer would also be dependent upon how intersecting the user base of both packages is going to be (which I assume it will intersect by quite a lot), and on how much the PyBOP core developers will be following up on changes to/in PyBaMM, so I guess there is no right answer here and everything will be evaluated on a case-by-case basis (the longer release cycle between both packages helps I guess!). As someone who does not use both packages that frequently and doesn't do battery modelling as a part of any institutional positions or an industrial job, I might be ringing the wrong bell here, so please feel free to redirect me.

BradyPlanden: Hi Agriya,

Great question, and one I've been thinking about. I think we should drop Python versions to follow PyBaMM (which then follows NumPy) within a release. We should definitely upper-pin PyBaMM, as I've noticed that there are still enough breaking changes from release to release to cause problems for PyBOP. I think this will mostly solve any Python version conflicts. 

I've also been thinking about how we support the range of PyBaMM versions, it will be challenging as we add/request functionality to PyBaMM to support PyBOP methods. For example, #252 was introduced because PyBaMM v23.5 doesn't have the Thevenin variables we need.

I think an update to the README is in order to add more granularity for users using PyBaMM standalone as well as PyBOP. Most of this should be solvable with proper virtualenvs, but I'm sure there will be edge cases and it will be good to have documentation to point users to.

Thanks for the review and the question!

PR #291, opened 2024-04-13
Add workflow to test PyBaMM develop branch nightly

### LINKED ISSUE #290: Test nightly against PyBaMM develop
### Feature description

Add a workflow to nightly test against PyBaMM `develop`. This will most likely be a stop-gap until the shared conda nightly builds with PyBaMM are completed (organised by @agriyakhetarpal). 

### Motivation

Allows us to capture how breaking changes from PyBaMM will affect PyBOP.

### Possible implementation

Update `pyproject.toml` within github action before running the test suite.

### Additional context

_No response_

BradyPlanden:
# Description

Adds a workflow to nightly test PyBOP against PyBaMM@develop. This should be used to capture how breaking changes from PyBaMM affect PyBOP.

## Issue reference
Fixes #290 

## Review

## Type of change
- [X] Testing: Additional tests to improve coverage or confirm functionality.

# Key checklist:

## Further checks:

### DIFF
--- a/.github/workflows/nightly_dependency_tests.yaml
+++ b/.github/workflows/nightly_dependency_tests.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -e .[all]
-          python -m pip uninstall pybamm
+          python -m pip uninstall -y pybamm
           python -m pip install git+https://github.com/pybamm-team/PyBaMM.git@develop
           python -m pip install pybtex
 
--- a/.github/workflows/nightly_dependency_tests.yaml
+++ b/.github/workflows/nightly_dependency_tests.yaml
@@ -27,62 +27,14 @@ jobs:
           python -m pip install pre-commit
           pre-commit run ruff
 
-  update-dependency:
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Checkout repository
-      uses: actions/checkout@v3
-
-    - name: Set up Python
-      uses: actions/setup-python@v3
-      with:
-        python-version: 3.12
-
-    # Update for pybamm develop
-    # with fix for pybtex optional dependency
-    - name: Update dependency version
-      run: |
-        python -m pip uninstall pybamm
-        python -m pip install pybtex
-        python -m pip install git+https://github.com/pybamm-team/PyBaMM.git@develop
-
-  integration_tests:
-    needs: [style, update-dependency]
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
-        python-version: ["3.12"]
-
-    name: Integration tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
-
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip nox
-
-      - name: Integration tests
-        run: |
-          nox -s integration
-
-  unit_tests:
-    needs: [style, update-dependency]
+  Tests:
+    needs: style
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-        exclude: # We run the coverage tests on Ubuntu with Python 3.12
-          - os: ubuntu-latest
-            python-version: "3.12"
         # Include MacOS M-series Runners
         include:
           - os: macos-14
@@ -92,7 +44,7 @@ jobs:
           - os: macos-14
             python-version: "3.12"
 
-    name: Unit tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
+    name: Test Suite (${{ matrix.os }} / Python ${{ matrix.python-version }})
 
     steps:
       - uses: actions/checkout@v4
@@ -102,65 +54,11 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip nox
+          python -m pip install -e .[all]
+          python -m pip uninstall pybamm
+          python -m pip install git+https://github.com/pybamm-team/PyBaMM.git@develop
+          python -m pip install pybtex
 
-      - name: Unit tests
+      - name: Run test suite
         run: |
-          nox -s unit
-
-  example_tests:
-    needs: [style, update-dependency]
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
-        python-version: ["3.12"]
-
-    name: Test examples (${{ matrix.os }} / Python ${{ matrix.python-version }})
-
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip nox
-
-      - name: Run examples
-        run: |
-          nox -s examples
-
-  # Runs only on Ubuntu with Python 3.12
-  check_coverage:
-    needs: [style, update-dependency]
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-    name: Coverage tests (ubuntu-latest / Python 3.12)
-
-    steps:
-      - name: Check out PyBOP repository
-        uses: actions/checkout@v4
-      - name: Set up Python 3.12
-        id: setup-python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.12
-          cache: 'pip'
-          cache-dependency-path: setup.py
-
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip nox
-      - name: Run coverage tests for Ubuntu with Python 3.12 and generate report
-        run: nox -s coverage
-
-      - name: Upload coverage report
-        uses: codecov/codecov-action@v4
-        with:
-          fail_ci_if_error: true
-          verbose: true
-          token: ${{ secrets.CODECOV_TOKEN }}
+          pytest --integration --unit --nbmake --examples
--- a/.github/workflows/nightly_dependency_tests.yaml
+++ b/.github/workflows/nightly_dependency_tests.yaml
@@ -54,7 +54,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+  schedule:
+    - cron: "0 23 * * *"
+
+concurrency:
+  # github.workflow: name of the workflow, so that we don't cancel other workflows
+  # github.event.pull_request.number || github.ref: pull request number or branch name if not a pull request
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  # Cancel in-progress runs when a new workflow with the same group name is triggered
+  # This avoids workflow runs on both pushes and PRs
+  cancel-in-progress: true
```
agriyakhetarpal, line 14: Since this won't run on pull requests or pushes but rather on schedule or manually, I will suggest using a different concurrency key to group the workflows (`github.ref` is available only for `pull_request:` and related events):

```suggestion
concurrency:
  # github.workflow: name of the workflow, so that we don't cancel other workflows
  # github.run_id || github.event_name: either the unique identifier for the job or the event that triggered it
  group: ${{ github.workflow }}-${{ github.run_id || github.event_name }}
  # Cancel in-progress runs when a new workflow with the same group name is triggered
  cancel-in-progress: true
```

I think it should work, because the scheduled events will be unaffected, and this should cancel duplicate `workflow_dispatch` runs. Otherwise, if it doesn't, we can drop the concurrency altogether later and cancel jobs manually.

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          python-version: 3.12
+
+      - name: Check formatting with pre-commit
+        run: |
+          python -m pip install pre-commit
+          pre-commit run ruff
+
+  Tests:
+    needs: style
```
agriyakhetarpal, line 32:
```suggestion
jobs:
  nightly_tests:
```

I feel we should drop the `style` checks from the nightly tests. If they fail, the nightly tests will fail to run.

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        # Include MacOS M-series Runners
+        include:
+          - os: macos-14
+            python-version: "3.10"
+          - os: macos-14
+            python-version: "3.11"
+          - os: macos-14
+            python-version: "3.12"
```
agriyakhetarpal, line 46: Question: do we need to test against all Python versions and all operating systems? It could soon become redundant, i.e., whether to drop support and bump the lower bound for a PyBaMM version is already being tested with the dynamic GitHub Actions matrix we created earlier, so we can use those tests to decide (we will be dropping just released versions of PyBaMM anyway, and testing develop-PyBaMM with something like Windows and Python 3.8 might not make sense. If it doesn't to you, too – we can consider keeping just `ubuntu-latest` + Python 3.12 and removing the matrix (and Python 3.13/dev if needed in a 1x2 matrix).

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install -e .[all,dev]
+          python -m pip uninstall -y pybamm
+          python -m pip install git+https://github.com/pybamm-team/PyBaMM.git@develop
+          python -m pip install pybtex
```
agriyakhetarpal, line 61: We fixed this in https://github.com/pybamm-team/PyBaMM/pull/3968 and merged recently, should be okay to remove now. :)

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install -e .[all,dev]
+          python -m pip uninstall -y pybamm
+          python -m pip install git+https://github.com/pybamm-team/PyBaMM.git@develop
```
agriyakhetarpal, line 60: Yet, to avoid PyBaMM's (sometimes occuring) import errors, something like this should work.

```suggestion
          python -m pip install "pybamm[all] @ git+https://github.com/pybamm-team/PyBaMM@develop"
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
...
+          python -m pip install pybtex
+
+      - name: Run unit tests
+        if: ${{ ! cancelled() }}
```
agriyakhetarpal, line 64: Do we still need this condition and in other places? (I thought you were using it to test the workflows in previous commits).

agriyakhetarpal (changes requested):

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,76 @@
+name: Nightly dependencies@develop
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 23 * * *"
```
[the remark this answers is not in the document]
agriyakhetarpal: It will most likely be weekly or fortnightly, and I will further recommend adding a job that needs this job and creates an issue for you. Here's how `xarray` started doing it, which is a bit complicated: https://github.com/pydata/xarray/pull/4583/ but it seems to work well: https://github.com/pydata/xarray/issues/8844

We can use a similar workflow based on this, or implement a simpler job that doesn't need to parse the logs and so on. and use existing third-party GitHub Actions that can do the same task (opening an issue, adding the workflow URL in the contents, etc.)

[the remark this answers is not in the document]
keewis: since I saw the cross-link: we've since moved on to using [xarray-contrib/issue-from-pytest-log](https://github.com/xarray-contrib/issue-from-pytest-log), which abstracts away a lot of the complex code and in general is a lot more robust.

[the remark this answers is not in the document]
agriyakhetarpal: That looks wonderful – thanks a lot, @keewis!

.github/workflows/nightly_dependency_tests.yaml
BradyPlanden: That's a good point, I'll update this to be `ubuntu-latest` + 3.12.

.github/workflows/nightly_dependency_tests.yaml
BradyPlanden: This is in place to ensure the job continues in the event one of these tasks fails. For example, without this addition, if the unit tests fail the remaining tasks get cancelled. The alternative would be to have all the `nox` session in a single task, but this gives us an quick way to see which tests failed (due to the task naming).

Open to other implementations if you have suggestions!

agriyakhetarpal: Ah, makes sense. How about a matrix to do this, though? It should be more elegant. I'll suggest some changes on how it can be done :)

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -18,16 +18,8 @@ jobs:
...
-        # Include MacOS M-series Runners
-        include:
-          - os: macos-14
-            python-version: "3.10"
-          - os: macos-14
-            python-version: "3.11"
-          - os: macos-14
-            python-version: "3.12"
+        os: [ubuntu-latest, macos-14]
+        python-version: ["3.12"]
```
agriyakhetarpal, line 22:
```suggestion
        os: [ubuntu-latest, macos-14]
        python-version: ["3.12"]
        suite: ["unit", "integration", "examples"]

File: .github/workflows/nightly_dependency_tests.yaml [hunk 9]
```diff
@@ -18,16 +18,8 @@ jobs:
...
 
     name: Test Suite (${{ matrix.os }} / Python ${{ matrix.python-version }})
```
agriyakhetarpal, line 24:
```suggestion
    name: Test Suite (${{ matrix.os }} / Python ${{ matrix.python-version }} / ${{ matrix.test-suite }})
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Run unit tests
+        if: ${{ ! cancelled() }}
+        run: |
+          python -m pytest --unit
+
+      - name: Run integration tests
+        if: ${{ ! cancelled() }}
+        run: |
+          python -m pytest --integration
+
```
agriyakhetarpal, line 47:
```suggestion
      - name: Run ${{ matrix.test-suite }} tests
        run: |
          if [ ${{ matrix.test-suite }} == "unit"]; then
              python -m pytest --unit
          elif [ ${{ matrix.test-suite }} == "integration"]; then
              python -m pytest --integration
          elif [ ${{ matrix.test-suite }} == "examples"]; then
              python -m pytest --nbmake --examples
          fi
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Run examples tests
+        if: ${{ ! cancelled() }}
+        run: |
+          python -m pytest --nbmake --examples
```
agriyakhetarpal, line 51:
```suggestion
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install -e .[all,dev]
+          python -m pip uninstall -y pybamm
+          python -m pip install "pybamm[all] @ git+https://github.com/pybamm-team/PyBaMM@develop"
+
+      - name: Run unit tests
+        if: ${{ ! cancelled() }}
+        run: |
+          python -m pytest --unit
```
agriyakhetarpal, line 41:
```suggestion
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install -e .[all,dev]
+          python -m pip uninstall -y pybamm
+          python -m pip install "pybamm[all] @ git+https://github.com/pybamm-team/PyBaMM@develop"
+
+      - name: Run unit tests
+        if: ${{ ! cancelled() }}
```
agriyakhetarpal, line 39:
```suggestion
```

agriyakhetarpal (commented):
I think GitHub is a bit buggy today so I'm not sure if the diffs I suggested are loading correctly. But I think these should be all of the necessary changes.

.github/workflows/nightly_dependency_tests.yaml
BradyPlanden: Looks good, thanks for the suggestions!

PR #340, opened 2024-06-03
Adds Monte Carlo Samplers

### LINKED ISSUE #6: Hamiltonian Monte Carlo (HMC)

BradyPlanden:
# Description

Adds Pints-based Monte Carlo samplers. The full list of composed samplers can be found [here](https://github.com/pybop-team/PyBOP/blob/a62a1269f12eacf7b54d8c4b2e205592fd0092c1/pybop/samplers/pints_samplers.py). 

This PR includes a base sampling class `BaseSampler`, of which `PintsBaseSampler` is a child. The Pints MCMC samplers are composed of PyBOP classes, which are constructed within `PintsBaseSampler`. This follows a similar structure to the `BasePintsOptimiser` class. A generic `MCMCSampler` class is included for use as a default, which is conceptually aligned with the `Optimisation` class.

Additions to the Priors class include the `JointLogPrior`.

Two additional PR's will be chained around this one. The order of merge should be:

- [X] [Transformations](https://github.com/pybop-team/PyBOP/pull/357)
- [ ] **This PR**
- [ ] Monte-carlo-plotting

## Issue reference
Fixes #6 

## Review

## Type of change
- [X] New Feature: A non-breaking change that adds new functionality.
- [X] Testing: Additional tests to improve coverage or confirm functionality.

# Key checklist:

## Further checks:


pybop/samplers/base_mcmc.py
```diff
@@ -0,0 +1,298 @@
...
+        self._initial_phase = self._samplers[0].needs_initial_phase()
+        if self._initial_phase:
+            self.set_initial_phase_iterations()
+
+        # Parallelisation (Might be able to move into parent class)
+        self._n_workers = 1
+        self.set_parallel(self._parallel)
+
+    def _apply_transformation(self, transformation):
+        # TODO: Implement transformation logic
```
BradyPlanden, line 126:
```suggestion
        # TODO: Implement transformation logic (alongside #357)
```

BradyPlanden:
This is ready for a first review. There is a bit more work to do on integration testing for the slower convergence samplers, as well as integrating #357. I believe this PR is in a good state to be integrated into develop, with subsequent PRs improving the design structure and robustness all around.

pybop/costs/_likelihoods.py
```diff
@@ -352,3 +352,99 @@ def _evaluateS1(self, inputs: Inputs) -> tuple[float, np.ndarray]:
         total_gradient = dl + prior_gradient
 
         return posterior, total_gradient
+
+
+class LogPosterior(BaseCost):
```
NicolaCourtier, line 357: This class should merge with and replace `MAP`.

pybop/parameters/priors.py
```diff
@@ -90,6 +122,98 @@ def rvs(self, size=1, random_state=None):
             loc=self.loc, scale=self.scale, size=size, random_state=random_state
         )
 
+    def __call__(self, x):
```
NicolaCourtier, line 125: I would suggest removing the `call`, `evaluate` and `_evaluate` functions from BasePrior because different values are needed across the code (e.g. `pdf`, `logpdf`), it is better to be specific about what is being evaluated here. The evaluateS1 function could be called `logpdfS1`

pybop/parameters/priors.py
```diff
@@ -199,3 +368,87 @@ def __init__(self, scale, loc=0, random_state=None):
...
+        -------
+        float
+            The value(s) of the first derivative at x.
+        """
+        log_pdf = self.__call__(x)
+        dlog_pdf = -1 / self.scale * np.ones_like(x)
+        return log_pdf, dlog_pdf
+
+
+class ComposedLogPrior(BasePrior):
```
NicolaCourtier, line 392: Let's rename and refactor `ComposedLogPrior` into `JointPrior` with `logpdf` and `logpdfS1` functions, analogous to `BasePrior`.

(In future, we could add functionality to marginalise the joint distribution to retrieve pdf's for the individual parameters - although the mapping would not be reversible if the parameters are correlated.)

pybop/parameters/priors.py
```diff
@@ -90,6 +122,98 @@ def rvs(self, size=1, random_state=None):
...
+            The point(s) at which to evaluate the first derivative.
+
+        Returns
+        -------
+        float
+            The value(s) of the first derivative at x.
+        """
+        raise NotImplementedError
+
+    def verify(self, x):
```
NicolaCourtier, line 207: I would not expect a dictionary to be passed here, so let's remove the verify function unless there is a particular reason for it?

pybop/parameters/priors.py
```diff
@@ -120,6 +244,10 @@ def sigma(self):
         """
         return self.scale
 
+    @property
+    def n_parameters(self):
```
NicolaCourtier, line 248: Can the number of parameters be different from 1? I think we should keep to individual parameter priors for now. In which case, let's move the `n_parameters` property to the joint prior only.

pybop/parameters/parameter.py
```diff
@@ -372,6 +372,12 @@ def get_sigma0(self) -> list:
 
         return sigma0
 
+    def priors(self) -> list:
```
NicolaCourtier, line 375: Should this be a `JointPrior` instead (see suggestion below)

pybop/samplers/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+
+from pybop import LogPosterior
+
+
+class BaseSampler:
+    """
+    Base class for Monte Carlo samplers.
+    """
+
+    def __init__(self, log_pdf, x0, cov0: Union[np.ndarray, float]):
```
NicolaCourtier, line 14: Update the input from `log_pdf` to `log_posterior` to make it clear that an instance of `LogPosterior` is expected here, not just a likelihood.

pybop/costs/_likelihoods.py
BradyPlanden: Agreed, but let's do this in a new PR, as it's out of scope on this one. I'll open an issue.

pybop/parameters/priors.py
BradyPlanden: Given the call stack to this point, it's likely to have this as a dict, as `parameters.verify(x)` within `cost.evaluate` provides a dict for the `cost.compute()`. In this situation, the `x` dict for the prior calculation in the LogPosterior class needs to be converted from dict to ndarray for the prior methods.

I think this is probably the best spot for the conversion, as the alternative would be to individually apply the conversion to each `cost.compute`, or refactor the prior methods to accept dicts. Do you have a preference?

pybop/parameters/parameter.py
BradyPlanden: Not sure if I follow on this one, are you suggesting it should be renamed `joint_priors` or something else?

pybop/parameters/priors.py
brosaplanella: The dictionary if statement is not currently checked by tests. If there is an easy test for this, it would be good to add. Otherwise, that might be a good argument towards removing the verify function (as in, if there is no easy test for that, it is unlikely that a user will inadvertently pass a dictionary).

pybop/parameters/parameter.py
brosaplanella: I think I have the same question as Nicola. Is (conceptually) this `priors` the same as the `JointPriors` later on or not necessarily?

pybop/samplers/base_pints_sampler.py
```diff
@@ -0,0 +1,291 @@
...
+    def _process_multi_chain(self):
+        reply = self._samplers[0].tell(self.fxs)
+        self._intermediate_step = reply is None
+        if reply:
+            ys, fys, accepted = reply
+            ys_store = np.array([self._inverse_transform(y, self._log_pdf) for y in ys])
+            if self._chains_in_memory:
+                self._samples[:, self._iteration] = ys_store
+            else:
+                self._samples = ys_store
```
brosaplanella, line 242: This line is not covered either

pybop/samplers/mcmc_sampler.py
```diff
@@ -0,0 +1,110 @@
...
+        ValueError
+            If the sampler could not be constructed due to an exception.
+        """
+
+        try:
+            self.sampler = sampler(log_pdf, chains, x0=x0, sigma0=cov0, **kwargs)
+        except Exception as e:
+            raise ValueError(
+                f"Sampler could not be constructed, raised an exception: {e}"
+            ) from e
```
brosaplanella, line 51: Is this really needed or could we just rely on the original Exception?

pybop/samplers/mcmc_sampler.py
```diff
@@ -0,0 +1,110 @@
...
+            The value to be set to the attribute.
+        """
+        if (
+            name in self.__dict__
+            or "sampler" not in self.__dict__
+            or not hasattr(self.sampler, name)
+        ):
+            object.__setattr__(self, name, value)
+        else:
+            setattr(self.sampler, name, value)
```
brosaplanella, line 110: This one is not covered either (though this one might be trickier, so happy to leave as is).

pybop/samplers/pints_samplers.py
brosaplanella, line 1: All the classes are defined very similarly. Could this be refactored to a single class called `PintsSampler` with an argument that is the sampler to import? That would drastically reduce code duplication.

tests/integration/test_monte_carlo.py
```diff
@@ -0,0 +1,167 @@
...
+    SliceDoublingMCMC,
+    SliceStepoutMCMC,
+    # Grad samplers
+    # NUTS,
+    # HamiltonianMCMC,
+    # MonomialGammaHamiltonianMCMC,
+    # RelativisticMCMC,
+    # SliceRankShrinkingMCMC,
+    # EmceeHammerMCMC,
+    # MALAMCMC,
```
brosaplanella, line 24: Why are these commented out? I would suggest to either remove, uncomment or explain why they are left as commented.

tests/integration/test_monte_carlo.py
```diff
@@ -0,0 +1,167 @@
...
+            HaarioBardenetACMC,
+            MetropolisRandomWalkMCMC,
+            PopulationMCMC,
+            RaoBlackwellACMC,
+            SliceDoublingMCMC,
+            SliceStepoutMCMC,
+        ],
+    )
+    # Samplers that either have along runtime, or converge slowly
+    # Need to assess how to perform integration tests with these samplers
```
brosaplanella, line 115: Actually that's what I meant earlier haha

brosaplanella (approved):
Looks good, just some small comments

pybop/samplers/pints_samplers.py
BradyPlanden: Ideally, yes. However, I have yet to find a nice implementation that enables IDE autocomplete, unfortunately. I'd like to keep the autocomplete functionality so users can start typing MCMC, AMC, NUTS, etc. and select the class or similar classes. I think to keep this, we have to construct the class within PyBOP with the corresponding name. I'm keen to find an alternative though!

brosaplanella: It's not as convenient as autocomplete, but if the sampler becomes a kwarg and the user provides something that doesn't make sense, one could use fuzzy string search to suggest a solution, similar to what PyBaMM does with parameters.

pybop/parameters/parameter.py
BradyPlanden: Gotcha, it's not necessarily a `JointPrior`, as conceptually it's a helper method to combine all the priors from the individual parameters into a list. This is used later within the `JointPrior`, but not exclusively.

pybop/samplers/pints_samplers.py
BradyPlanden: Thanks for the suggestion, that sounds like it would be quite helpful in the `MCMCSampler` class which takes a sampler as an argument. I think I'll open an issue to investigate this functionality for both the `MCMCSampler` and `Optimisation` classes.

PR #359, opened 2024-06-12
Pass Inputs dictionary

NicolaCourtier:
# Description

Aligning Inputs between problem, observer and model.

## Issue reference
Fixes #358.

## Review
Before you mark your PR as ready for review, please ensure that you've considered the following:
- Updated the [CHANGELOG.md](https://github.com/pybop-team/PyBOP/blob/develop/CHANGELOG.md) in reverse chronological order (newest at the top) with a concise description of the changes, including the PR number.
- Noted any breaking changes, including details on how it might impact existing functionality.

## Type of change
- [ ] New Feature: A non-breaking change that adds new functionality.
- [ ] Optimization: A code change that improves performance.
- [ ] Examples: A change to existing or additional examples.
- [ ] Bug Fix: A non-breaking change that addresses an issue.
- [ ] Documentation: Updates to documentation or new documentation for new features.
- [ ] Refactoring: Non-functional changes that improve the codebase.
- [ ] Style: Non-functional changes related to code style (formatting, naming, etc).
- [ ] Testing: Additional tests to improve coverage or confirm functionality.
- [ ] Other: (Insert description of change)

# Key checklist:

- [ ] No style issues: `$ pre-commit run` (or `$ nox -s pre-commit`) (see [CONTRIBUTING.md](https://github.com/pybop-team/PyBOP/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- [ ] All unit tests pass: `$ nox -s tests`
- [ ] The documentation builds: `$ nox -s doctest`

You can run integration tests, unit tests, and doctests together at once, using `$ nox -s quick`.

## Further checks:
- [ ] Code is well-commented, especially in complex or unclear areas.
- [ ] Added tests that prove my fix is effective or that my feature works.
- [ ] Checked that coverage remains or improves, and added tests if necessary to maintain or increase coverage.

Thank you for contributing to our project! Your efforts help us to deliver great software.

Issue #238, opened 2024-03-14
Allow multiple datasets & models at once


brosaplanella:
### Feature description

Allow the `FittingProblem` class to take lists of datasets and of models, which would streamline certain processes. The key idea is that when multiple models are present, they all use the same parameters (otherwise one can just write a for loop with N `FittingProblem`s).

Cases: 
- 1 model & 1 dataset: current case, just fit the model to the dataset.
- 1 model & N datasets: fit the model to minimise the cost function across datasets. A practical case for this would be when you run the same test on 3 different cells and want to fit the parameters.
- N models & N datasets: fit first model to first dataset, second to second and so on. Given that models are for specific operating conditions, an example of this would be if we wanted to fit a model to multiple C-rate experiments.
- N models & 1 dataset: I am still not convinced that this one would have any practical application, but it is the missing one so just listing it here.

### Motivation

Provided for each of the cases above.

### Possible implementation

Loop over and concatenate the error arrays before passing to cost function.

BradyPlanden:
Excellent thanks @brosaplanella, it looks like we were thinking along the same lines (#232). Happy to leave this one open as it has more background info.

To note, adding a weighting function to enable cost scaling across the datasets should be added at the same time as this functionality.

BradyPlanden:
Hi @MarkBlyth, just checking in to see how you are getting on with issue. If you are running into issues, or don't have the time to continue please let me know so that I can support or reassign.

MarkBlyth:
Hi Brady, I've been out the office for most of the past month so haven't had the chance to look at this yet. Still keen to work on it, and aiming to get started by the end of next week.

MarkBlyth:
Hi all, I've made a first pass, with code [here](https://github.com/MarkBlyth/PyBOP_develop/tree/%23238_multiple_datasets). Currently it will break the observer methods, hasn't been tested with multiple models, and has no unit tests, so it's very much work in progress.

The approach I've gone for is to make a DatasetCollection class, representing multiple datasets, and have the fitting_problem evaluate multiple models over multiple datasets, with the plotting and cost functions updated to accept multiple results at once. Given that this means making a large number of breaking changes to the code base, I was keen to get some feedback before I keep going with tests etc.

NicolaCourtier:
Sounds good, thanks @MarkBlyth! It would helpful at this stage to make a (draft) pull request so we can more easily see where you have made changes.

NicolaCourtier:
[...]

PR #364, opened 2024-06-14
Add MultiFittingProblem class and example

NicolaCourtier:
# Description

Add a new Problem class that accepts multiple fitting problems and returns one combined problem for passing to a cost function. This PR builds on and should be merged after #359.

## Issue reference
Fixes part of #238.

## Review

## Type of change

# Key checklist:

## Further checks:


examples/scripts/multi_fitting.py
```diff
@@ -0,0 +1,79 @@
...
+        "Current function [A]": values_2["Current [A]"].data,
+        "Voltage [V]": values_2["Voltage [V]"].data
+        + np.random.normal(0, sigma, len(values_2["Voltage [V]"].data)),
+    }
+)
+
+# Generate a problem for each dataset and combine into one
+problem_1 = pybop.FittingProblem(model_1, parameters, dataset_1)
+problem_2 = pybop.FittingProblem(model_2, parameters, dataset_2)
+problem = pybop.MultiFittingProblem(problem_list=[problem_1, problem_2])
```
brosaplanella, line 52: Context: see my comment in the inputs dictionary PR #359. I am concerned of what would happen if the parameters in `problem_1` and `problem_2` are not consistent. A solution would be for `MultiFittingProblem` to have same inputs as `FittingProblem` but take lists for models and datasets, while only allowing a single parameter list to be passed (plus additional arguments, such as weights). I am not sure if this would mess with other things though.

brosaplanella (approved):
Overall looks good, nice work! See my comment on the parameters.

pybop/problems/fitting_problem.py
```diff
@@ -134,3 +135,112 @@ def evaluateS1(self, inputs: Inputs):
...
+
+
+class MultiFittingProblem(BaseProblem):
+    """
+    Problem class for joining mulitple fitting problems.
+
+    Extends `FittingProblem` to multiple fitting problems.
+    """
+
+    def __init__(self, problem_list, weights=None):
```
BradyPlanden, line 147: I would covert this to unpack the args and form a list or dict with them. This is how we setup the `Parameters` class and provides a clean API :)

Something like:
```suggestion
    def __init__(self, *args, weights=None):
        self.problems = []
        for problem in args:
            self.problems.append(problem)
```

examples/scripts/multi_fitting.py
BradyPlanden, line 1: Can you change this to a notebook, or add a notebook in addition? I'm confident people will want to use this functionality and a notebook will add clarity.

pybop/problems/fitting_problem.py
```diff
@@ -134,3 +135,112 @@ def evaluateS1(self, inputs: Inputs):
...
+            {"Time [s]": np.asarray([]), "Combined signal": np.asarray([])}
+        )
+        for problem in self.problem_list:
+            for signal in problem.signal:
+                combined_dataset["Time [s]"] = np.concatenate(
+                    (combined_dataset["Time [s]"], problem._time_data)
+                )
+                combined_dataset["Combined signal"] = np.concatenate(
+                    (combined_dataset["Combined signal"], problem._target[signal])
+                )
```
BradyPlanden, line 167: I think these objects can be formed as a list then concatenated once at the end. `np.concatenate` is on the expensive side, and completing it once on a larger object can save overhead.

examples/scripts/multi_fitting.py
BradyPlanden: Just to add, an example fitting across multiple SOC's would be a great notebook showcasing this functionality.

examples/scripts/multi_fitting.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+
+import pybop
+
+# Parameter set and model definition
+parameter_set = pybop.ParameterSet.pybamm("Chen2020")
+model_1 = pybop.lithium_ion.SPM(parameter_set=parameter_set)
+model_2 = pybop.lithium_ion.SPM(parameter_set=parameter_set.copy())
```
BradyPlanden, line 8: I suspect we should implement a `copy()` method for the model class to simplify this.

tests/integration/test_model_experiment_changes.py
BradyPlanden, line 1: Let's add unit tests for the `MultiFittingProblem` class as well.

pybop/problems/fitting_problem.py
```diff
@@ -134,3 +135,112 @@ def evaluateS1(self, inputs: Inputs):
...
+                yi, dyi = problem.evaluateS1(problem_inputs)
+                y["Combined signal"] = np.concatenate(
+                    (y["Combined signal"], yi[signal])
+                )
+                if dy is None:
+                    dy = dyi
+                else:
+                    dy = np.concatenate((dy, dyi))
+
+        return (y, dy)
```
BradyPlanden, line 246: This completes a `evaluate()` for every output signal; however, we provide all signals for each `evaluate()`. I spent a bit of time thinking that the lower logic statement wasn't used, before realising it's a catch for the first iteration. I workshopped an alternative for you to consider:

```suggestion
        combined_signal = []
        all_derivatives = []
        
        for problem in self.problem_list:
            problem_inputs = problem.parameters.as_dict()
            signal_values, dyi = problem.evaluateS1(problem_inputs)
            
            # Collect signals and derivatives
            for signal in problem.signal:
                combined_signal.extend(signal_values[signal])
            all_derivatives.append(dyi)
        
        y = {"Combined signal": np.asarray(combined_signal)}
        dy = np.concatenate(all_derivatives) if all_derivatives else None
        
        return (y, dy)
```

pybop/problems/fitting_problem.py
```diff
@@ -134,3 +135,112 @@ def evaluateS1(self, inputs: Inputs):
...
+        y = {"Combined signal": np.asarray([])}
+        for problem in self.problem_list:
+            problem_inputs = problem.parameters.as_dict()
+            for signal in problem.signal:
+                yi = problem.evaluate(problem_inputs)
+                y["Combined signal"] = np.concatenate(
+                    (y["Combined signal"], yi[signal])
+                )
+
+        return y
```
BradyPlanden, line 208: Same as the `evaluateS1()` below.

```suggestion
        combined_signal = []
        
        for problem in self.problem_list:
            problem_inputs = problem.parameters.as_dict()
            signal_values = problem.evaluate(problem_inputs)
            
            # Collect signals
            for signal in problem.signal:
                combined_signal.extend(signal_values[signal])
        
        return {"Combined signal": np.asarray(combined_signal)}
```

BradyPlanden (commented):
Thanks for this @NicolaCourtier. I think it's coming along well, I've made a few comments for you to look at. As a more general thought, do you see any reason this couldn't be a `MultiProblem` class that composes multiple instances of`FittingProblem` or `DesignProblem`? Perhaps even combined instances of both (with an update to BaseOptimiser for `minimising` logic. This seems to be the logical next step, but we could do it now instead of refactoring this class later.

examples/scripts/multi_fitting.py
BradyPlanden: Right, we have one already: https://github.com/pybop-team/PyBOP/blob/4638a60f1bc00afef0bd2c8e179b7c17826289d4/pybop/models/base_model.py#L582

NicolaCourtier: At the moment, copying the model class does not seem to create a new `parameter_set`, which then causes problems if different parameters are later set as inputs. Do you think that `model.copy()` should point to the existing parameter set or create a copy of the parameter set?

NicolaCourtier: I added a `model.new_copy()` method to resolve this issue.

pybop/problems/fitting_problem.py
```diff
@@ -18,12 +19,25 @@ class FittingProblem(BaseProblem):
...
+    Additional Attributes
+    ---------------------
+    _dataset : dictionary
+        The dictionary from a Dataset object containing the signal keys and values to fit the model to.
+    _time_data : np.ndarray
+        The time points in the dataset.
+    n_time_data : int
+        The number of time points.
+    _target : np.ndarray
+        The target values of the signals.
```
brosaplanella, line 40: Here's a picky comment, but in PyBaMM we had a similar discussion a while ago: do we want to document methods and attributes meant for internal use? I believe it PyBaMM we decided not to, as users are not meant to use them, but there is good reasons to go either way.

brosaplanella (approved):
Looks good, thanks!

pybop/problems/fitting_problem.py
NicolaCourtier: That's an interesting decision! Let's include concise documentation for all attributes and methods in PyBOP. This information is vital for new developers.

examples/scripts/multi_fitting.py
NicolaCourtier: (This example requires #425)

NicolaCourtier: I have updated the example to fit across different initial SoC values. I think this is enough for now, let's open an issue to add a notebook as well.

NicolaCourtier:
Merge after #451

BradyPlanden (approved):
LGTM - thanks for the addition @NicolaCourtier!

PR #451, opened 2024-08-08
Merge model build and rebuild

### LINKED ISSUE #444: Refactor BaseModel.rebuild() and BaseModel.build()
### Feature description

Combine the methods with a clear docstring on when a new build is trigger compared to returning the currently build model. 

### Motivation

Simplifies the codebase and reduces maintance. Fundementally these methods can be combined without issue.

### Possible implementation

_No response_

### Additional context

_No response_

NicolaCourtier:
# Description

Refactor BaseModel so that rebuilding the model uses the same build functionality as building from scratch.

## Issue reference
Fixes #444

## Review

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybop/models/base_model.py
+++ b/pybop/models/base_model.py
@@ -136,7 +136,7 @@ class BaseModel:
         check_model : bool, optional
             If True, the model will be checked for correctness after construction.
         """
-        if parameters is not None:
+        if parameters is not None or inputs is not None:
             # Classify parameters and clear the model if rebuild required
             inputs = self.classify_parameters(parameters, inputs=inputs)
 
@@ -289,7 +289,7 @@ class BaseModel:
         self._disc = None
 
     def classify_parameters(
-        self, parameters: Parameters, inputs: Optional[Inputs] = None
+        self, parameters: Optional[Parameters] = None, inputs: Optional[Inputs] = None
     ):
         """
         Check for any 'rebuild_parameters' which require a model rebuild and
@@ -297,11 +297,15 @@ class BaseModel:
 
         Parameters
         ----------
-        parameters : pybop.Parameters
-            The input parameters.
+        parameters : Parameters, optional
+            The optimisation parameters. Defaults to None, meaning that the parameters
+            attribute is not modified.
+        inputs : Inputs, optional
+            The input parameters for the simulation (default: None).
         """
-        self.parameters = parameters or Parameters()
+        self.parameters = parameters or self.parameters
 
+        # Compile all parameters and inputs
         parameter_dictionary = self.parameters.as_dict()
         parameter_dictionary.update(inputs or {})
 
--- a/pybop/models/base_model.py
+++ b/pybop/models/base_model.py
@@ -256,11 +256,15 @@ class BaseModel:
             The dataset to be used in the model construction.
         """
         if "Current function [A]" in self._parameter_set.keys():
-            self._parameter_set["Current function [A]"] = pybamm.Interpolant(
-                dataset["Time [s]"],
-                dataset["Current function [A]"],
-                pybamm.t,
-            )
+            if "Current function [A]" not in self.parameters.keys():
+                current = pybamm.Interpolant(
+                    dataset["Time [s]"],
+                    dataset["Current function [A]"],
+                    pybamm.t,
+                )
+                # Update both the active and unprocessed parameter sets for consistency
+                self._parameter_set["Current function [A]"] = current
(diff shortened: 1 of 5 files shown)


pybop/models/base_model.py
```diff
@@ -238,37 +246,36 @@ def set_initial_state(self, initial_state: dict):
         # Use a copy of the updated default parameter set
         self._parameter_set = self._unprocessed_parameter_set.copy()
 
-    def set_params(self, rebuild: bool = False, dataset: Dataset = None):
+    def set_current_function(self, dataset: Union[Dataset, dict]):
```
BradyPlanden, line 249: Perhaps this should be a private method, since it wouldn't trigger a rebuild.

pybop/models/base_model.py
```diff
@@ -281,71 +288,26 @@ def clear(self):
...
+        Check for any 'rebuild_parameters' which require a model rebuild and
+        update the unprocessed_parameter_set if a rebuild is required.
 
         Parameters
         ----------
-        parameters : pybop.Parameters
-            The input parameters.
+        parameters : Parameters, optional
+            The optimisation parameters. Defaults to None, meaning that the parameters
+            attribute is not modified.
```
BradyPlanden, line 302: Just to make it clear that the internal parameters attribute is used instead.
```suggestion
        parameters : Parameters, optional
            The optimisation parameters. Defaults to None, resulting in the internal `pybop.Parameters` object to be used.
```

pybop/models/base_model.py
```diff
@@ -80,14 +80,10 @@ def __init__(
 
         Additional Attributes
         ---------------------
-        parameters : pybop.Parameters
+        pybamm_model : pybamm.BaseModel, optional
+            An instance of a PyBaMM model.
+        parameters : pybop.Parameters, optional
```
BradyPlanden, line 85: As these are internal attributes, I think the `optional` be removed here, correct?

```suggestion
        pybamm_model : pybamm.BaseModel
            An instance of a PyBaMM model.
        parameters : pybop.Parameters
```

pybop/models/base_model.py
```diff
@@ -238,37 +246,36 @@ def set_initial_state(self, initial_state: dict):
...
+        """
+        Update the input current function according to the data.
+
+        Parameters
+        ----------
+        dataset : pybop.Dataset or dict, optional
+            The dataset to be used in the model construction.
+        """
+        if "Current function [A]" in self._parameter_set.keys():
+            self._parameter_set["Current function [A]"] = pybamm.Interpolant(
```
BradyPlanden, line 259: It looks like the line below has been removed. Was this catch added elsewhere? It's functionality is for users wanting to fit a stationary current function from provided data, i.e. mark the current function as `"input"`.

```python
if "Current function [A]" not in self.parameters.keys():
```

BradyPlanden (approved):
Looks like a nice refactor, thanks @NicolaCourtier. A few suggestions for you to consider.

pybop/models/base_model.py
NicolaCourtier: Please can you expand- when would this require a rebuild?

pybop/models/base_model.py
NicolaCourtier: In this case, I don't think it would be poassible to pass a dataset?

pybop/models/base_model.py
BradyPlanden: Yeah no worries, here's the interaction I'm thinking of:

If I've built the model, then call this method, will the `_built_model` have the updated `_parameter_set` from this method? I don't think it would, but I may be missing the code flow here. If it wouldn't have the updated `_parameter_set` object, I think we want this method to be private: `_set_current_function`, so users know they shouldn't expect it to work correctly with a single call.

NicolaCourtier: Thanks, I've updated and added a test

pybop/models/base_model.py
NicolaCourtier: To be sure, I've added the check back in

pybop/models/base_model.py
BradyPlanden: Just to note, this works for `model.predict`, but won't for `model.simulate`. Anyway, something we can fix in the future if it pops up.

NicolaCourtier: I've just added model.simulate to the test. Please can you explain when it would not work?

NicolaCourtier: Seems to work, so I will go ahead and merge

PR #474, opened 2024-08-23
Fix transformations

### LINKED ISSUE #473: [Bug]: Inconsistencies in transformations
### Python Version

3.10

### Describe the bug

There are some inconsistencies in how the transformations are applied which mean that I received an error during the initialisation of an optimisation problem when employing the LogTransformation.

Setting up a PINTS optimiser with a LogTransformation and bounds on one or more of the parameters.

### Relevant log output

```shell
ValueError: Initial position must lie within given boundaries.
```

NicolaCourtier:
# Description

Move the cost evaluation error catches to `BaseOptimiser`, add `log_update()` and fix inconsistencies in the application of transformations.

## Issue reference
Fixes #473

## Review

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybop/costs/base_cost.py
+++ b/pybop/costs/base_cost.py
@@ -93,23 +93,16 @@ class BaseCost:
         self.parameters.update(values=list(inputs.values()))
         y, dy = None, None
 
-        try:
-            if self._has_separable_problem:
-                if calculate_grad is True:
-                    y, dy = self.problem.evaluateS1(self.problem.parameters.as_dict())
-                else:
-                    y = self.problem.evaluate(
-                        self.problem.parameters.as_dict(),
-                        update_capacity=self.update_capacity,
-                    )
+        if self._has_separable_problem:
+            if calculate_grad is True:
+                y, dy = self.problem.evaluateS1(self.problem.parameters.as_dict())
+            else:
+                y = self.problem.evaluate(
+                    self.problem.parameters.as_dict(),
+                    update_capacity=self.update_capacity,
+                )
 
-            return self.compute(y, dy=dy, calculate_grad=calculate_grad)
-
-        except NotImplementedError as e:
-            raise e
-
-        except Exception as e:
-            raise ValueError(f"Error in cost calculation: {e}") from e
+        return self.compute(y, dy=dy, calculate_grad=calculate_grad)
 
     def compute(self, y: dict, dy: ndarray, calculate_grad: bool = False):
         """
--- a/pybop/costs/base_cost.py
+++ b/pybop/costs/base_cost.py
@@ -35,7 +35,6 @@ class BaseCost:
 
     def __init__(self, problem: Optional[BaseProblem] = None):
         self.parameters = Parameters()
-        self.transformation = None
         self.problem = problem
         self.verbose = False
         self._has_separable_problem = False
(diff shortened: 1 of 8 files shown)


pybop/costs/base_cost.py
```diff
@@ -82,29 +85,20 @@ def __call__(self, inputs: Union[Inputs, list], calculate_grad: bool = False):
...
-                    y = self.problem.evaluate(
-                        self.problem.parameters.as_dict(),
-                        update_capacity=self.update_capacity,
-                    )
-
-            return self.compute(y, dy=dy, calculate_grad=calculate_grad)
-
-        except NotImplementedError as e:
-            raise e
+        if self._has_separable_problem:
```
BradyPlanden, line 92: I think the transformation should be applied within the cost class directly. This allows users to interact with the cost class with transformations if requested. One benefit of this is the user can create their own optimisation class, or investigate the transformed cost object directly. Was there a technical reason that this needed to be moved to the optimisation class?

pybop/optimisers/base_optimiser.py
```diff
@@ -199,6 +201,76 @@ def _run(self):
         """
         raise NotImplementedError
 
+    def cost_call(self, inputs: Union[Inputs, list], calculate_grad: bool = False):
```
BradyPlanden, line 204: This has a lot of duplicate code from the cost class, as such it could indicate that the application of the transformations in better suited there.

pybop/optimisers/base_optimiser.py
```diff
@@ -61,19 +62,18 @@ def __init__(
         self.parameters = Parameters()
         self.x0 = None
         self.bounds = None
-        self.sigma0 = 0.1
+        self.sigma0 = 0.02
```
BradyPlanden, line 65: Was there a specific condition that required this change? Shrinking the default `sigma0` reduces the explored space for the stochastic methods, which could lead to more local optimums for users that are not adjusting this value.

BradyPlanden (commented):
Thanks for the updates @NicolaCourtier! The transformation changes look good, I have a few comments on the change from applying transformations within the cost class to optimisation class.

pybop/optimisers/base_optimiser.py
NicolaCourtier: The optimiser.cost_call function is added to ensure that the error catches occur only when the cost is called by the optimiser. When a user calls the cost for an invalid parameter set, I believe it should return the error.

pybop/costs/base_cost.py
NicolaCourtier: At present, keeping transformations within the optimiser provides a cleaner and more reliable experience for the user. It also allows the user to pass custom cost functions (that are not pybop.cost objects) to the optimiser and take advantage of all the same functionality. I would vote to keep transformations in the optimiser class.

pybop/optimisers/base_optimiser.py
NicolaCourtier: The default value of 0.02 for just one of the Pints classes was not implemented correctly (the expected value was overwritten). The simplest change was to have all optimisers use the same default value.

NicolaCourtier:
I would recommend keeping properties, such as transformations, attached to a single base class for modularity (to reduce the chance of errors and unexpected behaviour). In this case, I think BaseOptimiser is conceptually the correct place for transformations.

BradyPlanden (approved):

PR #616, opened 2025-01-09
Refactors EIS methods as Mixin class to BaseModel

BradyPlanden:
# Description

This PR does the following:
- Refactors the EIS methods as a mixin class for BaseModel
- Combines `simulate()` and `simulateEIS()`, 
- General refactoring of `classify_parameters()`

## Issue reference
Fixes # (issue-number)

## Review

## Type of change

# Key checklist:

## Further checks:


martinjrobins (commented):
I am unsure of the utility of using a mixin here. Generally mixin classes are created to inject repeated code into many different classes, so creating one to only be used in one particular class seems anti-pattern.

I think it would make sense to not build the model in-place, since a given model can be built in multiple ways, but have the different uses of a given pybamm model generated different classes with the same interface. So have a `BaseModel`, a `BuiltModel` (which can run a simulation), and `BuiltModelEIS`, `BuildModelOther` derived from `BuiltModel`. Then you could have functions on BaseModel that builds and returns the other model classes. 

```python
class BaseModel:
   def build_eis(cost: BaseCost, params: Parameters) -> BuiltEISModel
   def build_with_geometric_params(cost: BaseCost, params: Parameters) -> BuildGeomModel
   def build_standard(cost: BaseCost, params: Parameters) -> BuiltStandardModel
```

BradyPlanden:
Thanks for the feedback @martinjrobins.

Interesting, I agree that the mixin here doesn't fully align with the "lego block" pattern used for mixins. The aim was to move these methods outside of the `BaseModel` file while still keeping them as class methods. I had an earlier implementation that moved them without retaining them as class methods, which involved passing the model to them directly in a more functional way, but this is a smaller refactor while still reducing the clutter within the `BaseModel` file.

```python
class BaseModel:
   def build_eis(cost: BaseCost, params: Parameters) -> BuiltEISModel
   def build_with_geometric_params(cost: BaseCost, params: Parameters) -> BuiltGeomModel
   def build_standard(cost: BaseCost, params: Parameters) -> BuiltStandardModel
```
Thanks for sharing this. I think stepping away from in-place building would be a general helpful. A few thoughts:

- Do you envision these classes declared within the above methods or just constructing instances?
- If I understand your proposal, this is essentially a builder pattern, right? So `BaseModel` would turn into a building class, and `BuiltModel` probably wouldn't inherit from `BaseModel`, as the build process doesn't need to be propagated to the built instances.

Structure-wise, I think the below is what you are proposing, correct? 

- BaseModel
- BuiltModel
    - BuiltGeomModel
    - BuiltEISModel
    - BuiltStandardModel

martinjrobins:
> * Do you envision these classes declared within the above methods or just constructing instances?

The classes (BuiltGeomModel, BuiltX) would be declared in sparate files, and the `build_x` functions could contruct instances of these.

> * If I understand your proposal, this is essentially a builder pattern, right? So `BaseModel` would turn into a building class, and `BuiltModel` probably wouldn't inherit from `BaseModel`, as the build process doesn't need to be propagated to the built instances.

yea, its a builder pattern but restricted in scope to the symbolic model and building a concrete model out of that. 

> 
> Structure-wise, I think the below is what you are proposing, correct?
> 
> * BaseModel
> * BuiltModel
>   * BuiltGeomModel
>   * BuiltEISModel
>   * BuiltStandardModel

Yup, that is it :)

PR #687, opened 2025-03-03
Adding OCP data LGM50 LT

noelhallemans:
OCP data of the LGM50LT cell provided by About:Energy for fitting measured impedance data.

# Description

Publishing the OCP data of the LGM50LT cell (provided by About:Energy) which was used in the paper for fitting measured impedance data.

## Type of change


BradyPlanden (approved):
LGTM - thanks!

PR #702, opened 2025-04-02
I700-problem

martinjrobins:
# Description

- Drafts out the problem classes
- adds a PybammPipeline class that acts as a wrapper around a pybamm simulation with rebuild functionality

### DIFF
--- a/pybop/problems/pybamm_problem.py
+++ b/pybop/problems/pybamm_problem.py
@@ -18,16 +18,23 @@ class PybammProblem(pybamm.Problem):
         self._pipeline = pybamm_pipeline
         self._cost_names = cost_names
 
+    def set_params(self, p: np.ndarray) -> None:
+        """
+        Sets the parameters for the simulation and cost function.
+        """
+        self.check_and_store_params(p)
+
+        # rebuild the pipeline (if needed)
+        self._pipeline.rebuild(self._params)
+
     def run(self) -> float:
         """
         Evaluates the underlying simulation and cost function using the
         parameters set in the previous call to `set_params`.
         """
         self.check_set_params_called()
-        self.check_params(self._params)
 
-        # rebuild the pipeline (if needed) and solve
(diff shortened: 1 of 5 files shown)


pybop/problems/python_problem.py
```diff
@@ -0,0 +1,46 @@
...
+import pybop
+
+
+class PythonProblem(pybop.Problem):
+    """
+    Defines a problem that uses a Python callable as the simulation + cost function to evaluate
+    """
+
+    def __init__(
+        self, f: Callable, f_with_sens: Callable, param_names: list[str] = None
```
BradyPlanden, line 14: Should `f` and `f_with_sens` be the same object?

pybop/problems/pybamm_problem.py
```diff
@@ -0,0 +1,65 @@
...
+        self,
+    ) -> tuple[float, np.ndarray]:
+        """
+        Evaluates the underlying simulation and cost function using the
+        parameters set in the previous call to `set_params`.
+        """
+        self.check_set_params_called()
+        self.check_params(self._params)
+
+        # rebuild the pipeline (if needed) and solve
```
BradyPlanden, line 49: I think we can remove this line, since we can't rebuild the pipeline and acquire sensitivities. Unless I'm missing something in the new pipeline interface.

BradyPlanden (approved):
Thanks @martinjrobins - this is looking good. Two small comments, but happy for this to be merged when you are ready

pybop/problems/python_problem.py
martinjrobins: I don't think so, the api we've for the problem classes splits these up into two different functions, so I thought we'd need two callables. But happy to discuss.

pybop/problems/pybamm_problem.py
martinjrobins: actually I think I messed this up, the rebuild should probably occur when set_params is called, I'll change this

PR #707, opened 2025-04-07
i701 ProblemBuilder

### LINKED ISSUE #701: Restructure - ProblemBuilder
ProblemBuilder:
Users create this first and use setters to set various aspects of a problem (e.g. costs, pybamm simulation, parameters). Finally a build function is called with creates a runner class (see below)

BradyPlanden:
# Description

Working PR for the ProblemBuilder class

## Issue reference
Fixes #701 

## Review

## Type of change
- [X] Refactoring: Non-functional changes that improve the codebase.

# Key checklist:

## Further checks:

### DIFF
--- a/pybop/_pybamm_pipeline.py
+++ b/pybop/_pybamm_pipeline.py
@@ -1,9 +1,9 @@
-from copy import copy
+from copy import copy, deepcopy
 
 import numpy as np
 import pybamm
 
-from pybop import Inputs
+from pybop import Inputs, Parameters
 
 
 class PybammPipeline:
@@ -25,6 +25,7 @@ class PybammPipeline:
         self,
         model: pybamm.BaseModel,
         parameter_values: pybamm.ParameterValues = None,
+        pybop_parameters: Parameters = None,
         solver: pybamm.BaseSolver = None,
         t_start: np.number = 0,
         t_end: np.number = 1,
@@ -49,9 +50,10 @@ class PybammPipeline:
             The parameters that will be used to rebuild the model. If None, the model will not be rebuilt.
         """
         self.requires_rebuild = False
-        self._parameter_names = []  # rebuild_parameters
         self._model = model
         self._parameter_values = parameter_values
+        self._pybop_parameters = pybop_parameters
+        self._parameter_names = pybop_parameters.keys()
         self._geometry = model.default_geometry
         self._methods = model.default_spatial_methods
         if solver is None:
@@ -71,17 +73,62 @@ class PybammPipeline:
         }
         self._submesh_types = model.default_submesh_types
         self._built_model = self._model
-        self._init_build()
+        self.requires_rebuild = self._determine_rebuild()
 
-    def _init_build(self) -> None:
+    def _determine_rebuild(self) -> bool:
         """
         The initial build process, useful to determine if
         rebuilding will be required.
         """
         model = self._model.new_copy()
-        self._parameter_values.process_geometry(self._geometry)
-        self._parameter_values.process_model(model)
-        self.requires_rebuild = self._parameters_require_rebuild(self._geometry)
+        parameter_values = self._parameter_values.copy()
+        geometry = deepcopy(self._geometry)
+
+        # Apply "[input]"
+        for parameter in self._pybop_parameters:
+            parameter_values.update({parameter.name: "[input]"})
+
+        parameter_values.process_geometry(geometry)
+        parameter_values.process_model(model)
+        requires_rebuild = self._parameters_require_rebuild(geometry)
+        if not requires_rebuild:
+            self._parameter_values = parameter_values
+        return requires_rebuild
+
+    def _parameters_require_rebuild(self, geometry) -> bool:
+        """
+        Checks whether the parameter values required a rebuild.
+        This is reimplemented with only the required functionality.
+        """
+        try:
+            # Credit: PyBaMM Team
+            for domain in geometry:
+                for spatial_variable, spatial_limits in geometry[domain].items():
+                    # process tab information if using 1 or 2D current collectors
+                    if spatial_variable == "tabs":
+                        for _, position_info in spatial_limits.items():
+                            for _, sym in position_info.items():
+                                self._process_and_check(sym)
+                    else:
+                        for _, sym in spatial_limits.items():
+                            self._process_and_check(sym)
+        except ValueError:
+            return True
+        return False
+
+    def _process_and_check(self, sym):
+        """
+        Process and check the geometry for each parameter.
+        This is reimplemented as geometric parameters are not
+        currently supported as InputParameters within Pybop.
+        Credit: PyBaMM Team
+        """
+        new_sym = self._parameter_values.process_symbol(sym)
+        leaves = new_sym.post_order(filter=lambda node: len(node.children) == 0)
+        for leaf in leaves:
+            if not isinstance(leaf, pybamm.Scalar):
+                raise ValueError("Geometry parameters must be Scalars")
+        return new_sym
 
     def rebuild(self, params: Inputs) -> None:
         """
@@ -98,7 +145,7 @@ class PybammPipeline:
                 f"Expected {len(self._parameter_names)} parameters, but got {len(params)}."
             )
 
-        for key, value in params:
+        for key, value in params.items():
             self._parameter_values[key] = value
 
         self.build()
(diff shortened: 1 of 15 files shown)


pybop/builders/pybamm.py
```diff
@@ -0,0 +1,109 @@
...
+        self._costs = []
+        self._cost_names = []
+        self._dataset = None
+        self._parameters = Parameters()
+        self._solver = []
+        self._parameter_values = []
+        self._rebuild_parameters = []
+        self._pipeline = []
+
+    def add_simulation(
```
martinjrobins, line 21: I'm a bit confused with the multiple simulations, how does this work in practice? Is each cost applied to each simulation and all added up? I thought we were going with only a single simulation for each build, and then multiple simulation would be combined later on via the python problem

pybop/builders/pybamm.py
```diff
@@ -0,0 +1,109 @@
...
+            self._pipeline.append(
+                PybammPipeline(
+                    model,
+                    param,
+                    self._solver[i],
+                    rebuild_parameters=self._rebuild_parameters,
+                )
+            )
+
+            return PybammProblem(
```
martinjrobins, line 87: this is returning after the 1st simulation is processed, is this correct?

pybop/builders/pybamm.py
```diff
@@ -0,0 +1,109 @@
...
+                pybamm_pipeline=self._pipeline,
+                param_names=self._parameters.keys(),
+                cost_names=self._cost_names,
+            )
+
+    def set_current_function(self) -> None:
```
martinjrobins, line 93: is this a private method? Doesn't seem like a user should call it

pybop/builders/pybamm.py
```diff
@@ -0,0 +1,109 @@
...
+    def _requires_rebuild(self, built_model: pybamm.BaseModel) -> bool:
+        solver = pybamm.CasadiSolver(mode="safe", atol=1e-6, rtol=1e-3)
+        try:
+            solver.solve(built_model, t_eval=[0, 1])
+        except pybamm.SolverError:  # Change to ValueError (i3649)
+            # If the solver fails, it indicates that the model needs to be rebuilt
+            return True
+        return False
+
+    def build(self) -> PybammProblem:
```
martinjrobins, line 52: first think probably should check that everything has been set/added correctly (probably doesn't make sense to add a simulation but not any costs

martinjrobins, line 52: returns only a single PybammProblem, but you have an add_simulation function that can add multiple simulations?

pybop/models/lithium_ion/basic_SPMe.py
```diff
@@ -600,8 +600,16 @@ def convert_physical_to_grouped_parameters(parameter_set):
     Cdl_n = parameter_set["Negative electrode double-layer capacity [F.m-2]"]
     m_p = 3.42e-6  # (A/m2)(m3/mol)**1.5
     m_n = 6.48e-7  # (A/m2)(m3/mol)**1.5
-    sigma_p = parameter_set["Positive electrode conductivity [S.m-1]"]
-    sigma_n = parameter_set["Negative electrode conductivity [S.m-1]"]
+    sigma_p = (
```
martinjrobins, line 603: not sure what these changes are for?

martinjrobins (changes requested):
looks great, thanks @BradyPlanden, a few questions below. I thought that each builder should only build a single pybamm simulation, but you have an add_simulation function?

pybop/builders/pybamm.py
BradyPlanden: Hi Martin, thanks for taking a look at this. In terms of single vs multi simulation, I wanted to investigate how we may do this within the builder/problem itself. During the development day, we had split this into two separate classes, but I think we can maintain them better as a single class with the `add_simulation` API, which (I think) is quite appealing. In terms of weighting, I've added an optional args to `add_simulation` and `add_cost` that allow the user to weight the simulation and costs relatively. This is another improvement on our current API as the interaction with weighting simulations and costs are now consistent. Let me know your thoughts!

pybop/builders/pybamm.py
BradyPlanden: Yes, I've updated the PybammProblem class to include multiple pipelines. The problem.run() method now returns the resulting simulation/cost weighted output.

pybop/_pybamm_pipeline.py
```diff
@@ -70,40 +71,61 @@ def __init__(
...
-        model = self._model.deep_copy()
-        geometry = self._geometry.deep_copy()
-        self._parameters.process_geometry(geometry)
-        self._parameters.process_model(model)
+        for key, value in params:
+            self._parameter_values[key] = value
+
+        self.build()
+
+    def build(self) -> None:
```
martinjrobins, line 106: do we need a separate (public) build function? Seems like this should all go in `_init_build` since you only do it once on construction?

pybop/builders/pybamm.py
martinjrobins: what I worry about is that it is a bit restrictive and adds complexity. Restrictive in that a user might not want to add all costs to all models, they might want to add `n` costs, one to each of `n` models, or something else entirely. Adds complexity in that we're already planning on implementing a python builder and a pybamm builder, and the multiple simulation case can already be done within this framework, so adding complexity to the builder class is not needed (as far as I can see, I might be missing something)

pybop/costs/pybamm_cost.py
```diff
@@ -0,0 +1,41 @@
...
+    A SumSquaredError cost implementation within Pybamm.
+    """
+
+    @staticmethod
+    def variable_name():
+        return "SumSquaredError"
+
+    @staticmethod
+    def variable_expression():
+        return pybamm.ExplicitTimeIntegral(
```
martinjrobins, line 34: you want to use `pybamm.DiscreteTimeData` and `pybamm.DiscreteTimeSum`, see the test linked below for an example of use:
https://github.com/pybamm-team/PyBaMM/blob/cb17226225a236ee542664c353d1a7a3219ae39d/tests/unit/test_expression_tree/test_unary_operators.py#L762

pybop/_pybamm_pipeline.py
```diff
@@ -112,16 +134,63 @@ def built_model(self):
...
+        """
+        new_sym = self._parameter_values.process_symbol(sym)
+        leaves = new_sym.post_order(filter=lambda node: len(node.children) == 0)
+        for leaf in leaves:
+            if not isinstance(leaf, pybamm.Scalar):
+                raise ValueError("Geometry parameters must be Scalars")
+        return new_sym
+
+    def solve(
+        self, inputs: Inputs = None, calculate_sensitivities: bool = False
```
martinjrobins, line 179: I don't think `inputs` should be an arg here, these are already given in `rebuild`. `solve` should just solve for the inputs last given in `rebuild`

martinjrobins (changes requested):
Thanks @BradyPlanden, I'm still not convinced on the multiple simulations in a single pybamm problem, why is this needed? In regards to the cost, you are using the incorrect pybamm classes, I've linked to the correct ones below

pybop/_pybamm_pipeline.py
BradyPlanden: I believe so, yes. The `_init_build` only builds enough so that we can determine if the simulation will require rebuilding or not via `_parameters_require_rebuild`. Essentially we need to know whether of not to set the corresponding pybamm parameters to `input` or not before going through the full build.

pybop/_pybamm_pipeline.py
BradyPlanden: That's essentially what's happening here. The `set_params` method within the `PybammProblem` updates the `self._params` values and when we call `problem.run()` those values are passed to the pipeline. Since the pipeline is composed within the problem class, it doesn't have access to the `_params` of the parent. Are you suggesting we store the `_params` within the pipeline instead?

pybop/_pybamm_pipeline.py
martinjrobins: I'm mainly looking at the complicated back and forth between the builder class and the pipeline class in terms of rebuilding:
```python
            # Construct the pipeline
            pipeline = PybammPipeline(
                model,
                param,
                self._solver[i],
            )

            if not pipeline.requires_rebuild:
                # set input parameters
                for parameter in self._pybop_parameters:
                    param.update({parameter.name: "[input]"})

            # Build the pipeline, determine if the parameters require rebuilding
            pipeline.build()

            # Add to the parameter names attr if rebuild required
            if pipeline.requires_rebuild:
                pipeline.parameter_names = self._pybop_parameters.keys()
            self._pipeline.append(pipeline)
```
As it is, the builder class has to be aware about rebuilding and take this into account when it constructs the pipeline. I think it would be simpler if only the pipeline needs to know about rebuilding and handles this itself within its own constructor. All it would require is that you pass in the names of the pybop parameters to the pipeline constructor, then it can handle all the rebuilding setup stuff

pybop/_pybamm_pipeline.py
martinjrobins: yea I think we have to (store the params in preparation for the solve). `PybammPipeline` is not aware of anything going on in the problem class. As far as its concerned it has rebuilt the model with a certain set of inputs that was passed into `rebuild`. All `solve` should do is to run the model with those inputs. If you pass `inputs` into `solve`, you have no way of knowing if that `inputs` is the same as the `inputs` passed to `rebuild`.

BradyPlanden:
Thanks for the reviews @martinjrobins - this is now ready for another pass. I've left `_init_build` and `build` as separate methods within the pipeline as I'd like to reduce the logic completed within the `__init__`. This gives the pipeline rebuild information on construction, without completing a full build until it is called. Within the builder class this doesn't matter much, but minimising potential failures on pipeline initialisation seems worthwhile.

martinjrobins (approved):
Thanks @BradyPlanden, I'm happy for this to be merged whenever you are ready. I'm agree that `init_build` and `build` should be separate methods to keep the functions smallish, but I still think `build` should be a private method that is just called by the `__init__` and no-one else. Presumably there will never be a case where you want to construct a pipeline and then *not* call `build` straightaway. But its a small thing so happy for it to be merged

PR #717, opened 2025-05-06
Add EIS Builder

### LINKED ISSUE #701: Restructure - ProblemBuilder

BradyPlanden:
# Description

- Adds an EISBuilder and EISProblem 
- Adds `initial_state` functionality to the the PybammPipeline
- Adds python costs with domain weighting

## Issue reference
Fixes #701 

## Review

## Type of change

# Key checklist:

## Further checks:

### DIFF
--- a/pybop/_pybamm_eis_pipeline.py
+++ b/pybop/_pybamm_eis_pipeline.py
@@ -1,3 +1,5 @@
+import warnings
+
 import casadi
 import numpy as np
 import pybamm
@@ -10,7 +12,7 @@ from pybop._pybamm_pipeline import PybammPipeline
 
 class PybammEISPipeline(PybammPipeline):
     """
-    A class to build a PyBaMM pipeline for a given model and data, and run the resultant simulation.
+    A class to build an EIS PyBaMM pipeline for a given model and data, and run the resultant simulation.
 
     There are two contexts in which this class can be used:
     1. A pybamm model needs to be built once, and then run multiple times with different input parameters
@@ -26,14 +28,14 @@ class PybammEISPipeline(PybammPipeline):
     def __init__(
         self,
         model: pybamm.BaseModel,
+        f_eval: list,
         parameter_values: pybamm.ParameterValues = None,
         pybop_parameters: Parameters = None,
         solver: pybamm.BaseSolver = None,
-        f_eval: list = None,
         var_pts: dict = None,
     ):
         """
-        Arguments
+        Parameters
         ---------
         model : pybamm.BaseModel
             The PyBaMM model to be used.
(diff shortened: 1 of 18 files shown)


pybop/_pybamm_eis_pipeline.py
```diff
@@ -10,7 +12,7 @@
 
 class PybammEISPipeline(PybammPipeline):
```
martinjrobins, line 13: I don't think you should inherit from `PybammPipeline`, since you are changing the API (the solve function returns a different type). In a normal inheritance situation, you should be able to treat an instance of the child class (PybammEISPipeline) as if its an instance of the parent class (PybammPipeline). If you want to reuse code from `PybammPipeline` then just create an instance of this internaly and use it (ie. composition)

pybop/_pybamm_eis_pipeline.py
```diff
@@ -151,40 +177,51 @@ def initialise_eis_pipeline(self):
...
         A = 1.0j * 2 * np.pi * frequency * self.M - self._jac
 
         # Solve the system
         x = spsolve(A, self.b)
 
-        # Calculate the impedance
+        # Calculate the impedance (voltage / current)
         return -x[-2] / x[-1]
 
     def solve(self, calculate_sensitivities: bool = False) -> np.ndarray:
```
martinjrobins, line 204: API is different here to base class (return type is different).

pybop/_pybamm_eis_pipeline.py
```diff
@@ -151,40 +177,51 @@ def initialise_eis_pipeline(self):
...
 
         Returns
         -------
-        sol : dictionary
-            A dictionary containing the impedance values with corresponding frequencies.
+        np.ndarray
+            Complex array containing the impedance values with corresponding frequencies.
         """
+        if calculate_sensitivities:
+            warnings.warn(
```
martinjrobins, line 220: ideally I think the api should be

```python
def solve(self) -> np.ndarray:
```

so its impossible to mis-use this class and try to caculate the sensitivities

martinjrobins (changes requested):
looks good, thanks @BradyPlanden. However I don't think inheriting from `PybammPipeline` is the way to go here as the API seems quite different between the two classes

pybop/_pybamm_eis_pipeline.py
BradyPlanden: That's a fair point, I had modified the solve return type but it originally it matches the `PybammPipeline` type. I'm a bit tore on whether this should be an inherit from PybammPipeline or compose it. It definitely fits the context of 'is' a `PybammPipeline` with specific extensions, on the other it could `have` a `PybammPipeline`. Composing it will be a bit clunky I think, since within the `EISProblem` class we will have `self._pipeline.pybamm_pipeline`. I'm leaning toward restoring the solve API so that it aligns with `PybammPipeline` and continuing with inheritance. Thoughts?

pybop/_pybamm_eis_pipeline.py
BradyPlanden: At the moment this isn't supported, but I think we should aim to support gradient information for the EIS problems.

martinjrobins: ok, fair enough

PR #740, opened 2025-06-12
refactor: samplers

### LINKED ISSUE #722: Samplers:refactor to work with new problem class

martinjrobins:
# Description

refactor samplers to new design

## Issue reference
Fixes #722


pybop/samplers/base_pints_sampler.py
```diff
@@ -253,51 +253,45 @@ def _create_evaluator(self):
...
+            return pints.ParallelEvaluator(f, n_workers=self.options.n_workers)
 
         # Construct a dict for various return types
-        evaluator_map = {False: SequentialEvaluator, True: MultiSequentialEvaluator}
-        return evaluator_map[self._multi_log_pdf](f)
+        evaluator_map = {
+            False: pints.SequentialEvaluator,
+            True: pints.MultiSequentialEvaluator,
+        }
+        return evaluator_map[False](fun)
```
BradyPlanden, line 278: It looks like the evaluator map dictionary can be removed, as it's now fixed to `SequentialEvaluator`.

pybop/samplers/base_pints_sampler.py
```diff
@@ -124,7 +124,7 @@ def _initialise_chain_processor(self):
         else:
             self._chain_processor = MultiChainProcessor(self)
 
-    def run(self) -> Optional[np.ndarray]:
+    def run(self) -> np.ndarray:
```
BradyPlanden, line 127: This is still an optional output, depending on the attribute `chains_in_memory`

BradyPlanden (commented):
Looks good, thanks @martinjrobins! Just two small comments

pybop/samplers/base_pints_sampler.py
BradyPlanden: The usecase is when a user only wants to store the chains offline so enables `chain_files`. In this situation, `_chains_in_memory` can be `False` so the chains aren't also stored as an object in the python environment. Tbh, I'd be open to removing this functionality, as I don't believe it's a path used heavily by users and does complicate the samplers a decent amount. This could be a follow on PR though.

BradyPlanden (approved):
LGTM, thanks @martinjrobins

PR #741, opened 2025-06-18
refactor: fix classifier and dataset tests

martinjrobins:
# Description

Part of the refactor, fixes classifier and dataset unit tests


BradyPlanden (approved):
Thanks @martinjrobins, LGTM

Issue #734, opened 2025-05-30
Handling of optimisation results is unclear


katielukow-AE:
### Feature description

It would be great if the examples were more clear on how to handle the resulting output from the optimisation class. For example, how to access the PyBaMM solution with the identified parameter set. 

### Motivation

To provide insight on how to handle the resulting identified parameters outside of built-in plots

NicolaCourtier:
Hi @katielukow-AE, thanks for opening this issue! The code has changed quite a lot since, so I wanted to check what (if any) improvement you would still like to see.

The typical workflow now involves creating a PyBaMM `model`, `simulator` and optimisation `problem`. Running an optimiser or sampler returns a `result` for the given problem, which includes the identified `best_inputs`.

The result can be used for further investigation, for example:

1. Reproduce the optimised solution using `simulator.solve(inputs=result.best_inputs)`

2. Reproduce the cost value (and sensitivities) using `problem.evaluate(inputs=result.best_inputs)`

3. Continue with PyBaMM using `parameter_values.update(result.best_inputs)` then e.g. create a new simulation with `pybamm.Simulation(model, parameter_values=parameter_values)`

Do these examples resolve the issue?

PR #765, opened 2025-07-11
Restructure examplestests

### LINKED ISSUE #724: examples: refactor examples to use new builder class

MarkBlyth:
# Description

Example scripts need updating to the new PyBOP API, as per #724. I've made a start on this with the tau_reparameterised_ecm.py example, since that's the sort of problems I do most of my PyBOPping on. The other examples will all need updating, but this PR will hopefully provide a starting point for how the new API is used, and what the other examples should look like.

## Issue reference
Initial step towards #724 

## Review
- Haven't updated the change log, since there's still plenty more to do to fix #724.
- No breaking changes


MarkBlyth:
Some user-end thoughts on the new API. There's quite a lot of suggestions in here and they're all cosmetic rather than functional, but would hopefully make the PyBOP interface a bit more intuitive for new users. Can open these as issues if there's interest / agreement on any of these points.

### Parameters

I don't particularly like the current parameters setup. Consider the snippet from lines 68--91 and 117--128 of tau_reparameterised_ecm.py on this commit. We have...

```
parameters = pybamm.ParameterValues(
    {

       ...

        "R0 [Ohm]": 0.0008,
        "R1 [Ohm]": 0.001,
        "C1 [F]": 10150,
        "Element-1 initial overpotential [V]": 0,
    }
)

...

builder.set_simulation(model, parameter_values=parameters)

builder.add_parameter(
    pybop.Parameter(
        "R0 [Ohm]",
        initial_value=0.0008,
    )
)
builder.add_parameter(
    pybop.Parameter(
        "R1 [Ohm]",
        initial_value=0.001,
    )
)
builder.add_parameter(
    pybop.Parameter(
        "tau1 [s]",
        initial_value=10.15,
    )
)
```

It seems odd to define the model parameters twice --- once when setting up the simulation, and again when setting up the fitting problem. To my mind, parameters should be split into those we know already, which should be passed when setting up the simulation, and those we're trying to fit, which are added to the builder when setting up the fitting problem. This is only a minor change, but would make for a simpler, clearer API. Implementing that should be simple: all it should need is for PyBOP to call update on the simulation parameters when building the model.


### `builder` API

Lines 109--132 of tau_reparameterised_ecm.py on this commit:

```
builder = pybop.builders.Pybamm()
builder.set_dataset(fitting_data)
builder.set_simulation(model, parameter_values=parameters)
builder.add_parameter(
    pybop.Parameter(
        "R0 [Ohm]",
        initial_value=0.0008,
    )
)
builder.add_parameter(
    pybop.Parameter(
        "R1 [Ohm]",
        initial_value=0.001,
    )
)
builder.add_parameter(
    pybop.Parameter(
        "tau1 [s]",
        initial_value=10.15,
    )
)
builder.add_cost(
    pybop.costs.pybamm.SumSquaredError("Voltage [V]", "Voltage [V]"),
)
problem = builder.build()
```

This seems like a fairly clunky way to set up the problem. I think a functional interface would be neater, which might look something along the lines of

```
known_parameters =  {
        "Initial SoC": 0.75,
        "Initial temperature [K]": 25,
        "Cell capacity [A.h]": 5,
       ...
}
fitting_params = {"R0 [Ohm]" : 0.0008, "R1 [Ohm]: 0.001, "tau1 [s]": 10.15}
problem = pybop.Problem(model, known_parameters, pybop.costs.SumSquaredError)
opti_soln = problem.optimise(fitting_params, target_data, optimiser)
```
...where `target_data` is the `Dataset` to fit to, and `optimiser` is the choice of optimisers to use. In this case, `Problem` could be a simple helper function that invokes the correct builder for the given `type(model)`, and runs all the assembly and setting behind the scenes.

### Cost functions

Currently, cost functions are set up as...

```
builder.add_cost(
    pybop.costs.pybamm.SumSquaredError("Voltage [V]", "Voltage [V]"),
)
```

This seems a bit tightly coupled to the problem-domain. I'd think of a cost function as taking two sets of data --- one for expected, one for reality --- and returning a distance metric between them. It wouldn't need to know what those data mean or where they came from. An alternative could use the `Dataset` to record what our reference signal is (since that's already holding the reference signal itself). Eg...

```
fitting_data = pybop.Dataset(
    {
        "Time [s]": sol.t,
        "Current function [A]": sol["Current [A]"].data,
        "Voltage [V]": corrupt_values,
        "Reference data": "Voltage [V]",
    }
)
```

and have the cost function itself as any valid callable...

```
class BaseCostFunc(abc.ABC):
    @abstractmethod
    def __call__(self, expectation: np.ndarray, reality: np.ndarray) -> float:
        pass
```

The pybop `Problem` can pull `"Reference data"` out of each simulation result, and pass the relevant data to the cost function.


### Scipy optimizers

This one's really minor. Current setup:

```
optimiser = pybop.SciPyMinimize(problem, pybop.ScipyMinimizeOptions(method="Nelder-Mead"))
```

Do the optimiser options need to be a dataclass? It seems more pythonic to pass them as kwargs, and let scipy do any argument checking itself, rather than building a separate dataclass to do that for us.

Also, we have `SciPyMinimize` and `ScipyMinimizeOptions`. Maybe change to `ScipyMinimize` for consistency? (Just nitpicking here!)

### Richer optimisation outputs

When we call 
```fit = optimiser.run()```
the returned solution could include a dict or PyBaMM parameter set of some sort, storing the fitted parameters. This is basically so that users don't have to remember which parameters `fit.x[0]` etc. correspond to. Since we set up the problem as
```
builder.add_parameter(
    pybop.Parameter(
        "tau1 [s]",
        initial_value=10.15,
    )
)
```
it would be nice to get back `fit.pars = {"tau1 [s]": 20.0}` or similar.

Further 'nice extra feature': getting a simulation straight from the optimised result. Instead of...
```
problem.set_params(fit.x)
fitsol = problem.pipeline.solve()
```
how about...
```
fitsol = fit.simulate()
```

**Comment by BradyPlanden** [comment]
Thanks @MarkBlyth! Really great to hear your feedback, there are a few reasons for the sharp-edges you've found. I'm going to write them out below, with the (hopeful) goal that parts of this comment can be re-used in the documentation. 

### Parameters
I agree, it's a bit clunky to construct the `pybop.parameter` before passing to the builder. An update to the builder so that it constructs the parameter object during the `add_parameter` method is welcome. However, the reason we moved away from this was to be more explicit with the arguments passed. For example, it would preferable to have named arguments that match the parameter class `__init__` instead of a dict object or similar. In this situation, I think we would want to commit to either `Parameter` objects or `Parameter` construction objects, not both. 

> It seems odd to define the model parameters twice --- once when setting up the simulation, and again when setting up the fitting problem. To my mind, parameters should be split into those we know already, which should be passed when setting up the simulation, and those we're trying to fit, which are added to the builder when setting up the fitting problem. 

We don't rely on the `pybamm.Simulation` class in the restructure architecture. In these examples it's available since we are working with synthetic data, but under experimental fitting the builder would be the first point of passing parameter values.

### Builder

> This seems like a fairly clunky way to set up the problem. I think a functional interface would be neater, which might look something along the lines of

Right, yes this is similar to current PyBOP. There is room for us to build wrapping methods for builders (see MultiFittingBuilder), but I would advocate that the verbosity of the builder is actually linked with the complexity of the underlying optimisation task that PyBOP performs. By abstracting these decisions/interfaces away from the user, we have a much harder job maintaining the underlying codebase and the users' are mostly unaware on how to extend or modify the optimisation task for their problem. I think the builder structure does a better job balancing verbosity with functionality, at least compared to the historic pybop interface. To reduce verbosity, chaining will be available:

```python
builder = (
    pybop.builders.Pybamm()
    .set_dataset(fitting_data)
    .set_simulation(model, parameter_values=parameters)
    .add_parameter(pybop.Parameter("R0 [Ohm]", initial_value=0.0008))
    .add_cost(pybop.costs.pybamm.SumSquaredError("Voltage [V]", "Voltage [V]"))
)

problem = builder.build()
```
 *Note: if you would like to update the builders to ensure they all have this functionality that would be appreciated!*

Hopefully this makes sense, in general, there are absolutely improvements we can make to the builder, but I think providing this interface gives users a very clear central API to build a mental model around. My intuition is that smaller number of wrapper classes we need to create the better, as each level requires documentation, maintenance, and user commitment. 

### Cost functions
> This seems a bit tightly coupled to the  problem-domain.

Yes, they are. This is due to the way we embedded the cost as a node in the PyBaMM expression tree. The current import is quite verbose, and there is an argument to be made for the PyBaMM costs becoming the default so we have something like: ` pybop.costs.SumSquaredError("Voltage [V]")` for the majority of the use-cases.

> An alternative could use the Dataset to record what our reference signal is (since that's already holding the reference signal itself).

I think we can accomplish this by checking if the provided name is within the dataset if only a single value is passed. I.e. `pybop.costs.pybamm.SumSquaredError("Voltage [V]")`. An error could be raised if the name is not within the `Dataset`, during the `builder.build()` with information on providing the correct name to the cost class.

### Scipy optimizers
As a single line it is quite verbose, but I it does really improve clarity to both users and developers. Kwargs are notoriously difficult to reason over and conveying options to users then needs to be done via documentation which is almost always lacking or out of date 😄. The options dataclass ensures options are very clearly stated (and show up in IDE's). I do think multi-lining does improve it:
```python
options = pybop.ScipyMinimizeOptions(method="Nelder-Mead", maxiter=100)
optimiser = pybop.SciPyMinimize(problem, options=options )
``` 

### Richer optimisation outputs
Yes, I think there is work to be done within this class (see #734 also). Given the restructure, which now copies the provided objects to avoid changing the users model, parameter, etc. I think we should provide the complete `pybamm.ParameterValues` object with the optimised/sampled parameters for users to then benefit from PyBaMM's visualisation classes. Ideally, PyBOP completes the optimisation/sampling and provides the complete parameters so users can continue on their way.

Massive thanks for going through this and providing feedback!

PR #787, opened 2025-08-11
bug: fix optimiser logging after multiprocessing changes

martinjrobins:
# Description

After the multiprocessing updates (problem.run returns a numpy array rather than float) there were a few optimisers that did not store the correct types in the logger object, breaking some plots. This fixes these, and adds tests to ensure that the types stored in the logger are consistant


BradyPlanden (approved):
Thanks @martinjrobins — I've aligned both of our changes where possible. Feel free to modify or merge as is!

PR #802, opened 2025-09-01
Remove set_params

NicolaCourtier:
# Description

Remove the `set_params` functionality and just pass the inputs to the solution method.

## Type of change

# Important checks:

### DIFF
--- a/pybop/pipelines/_pybamm_pipeline.py
+++ b/pybop/pipelines/_pybamm_pipeline.py
@@ -1,17 +1,9 @@
-from copy import copy, deepcopy
-from enum import Enum
+from copy import deepcopy
 
 import numpy as np
 import pybamm
 
 from pybop import FailedSolution, Inputs, Parameters, RecommendedSolver
-
-
-class OperatingMode(Enum):
-    """Enum-like class for operating modes."""
-
-    WITHOUT_EXPERIMENT = "without experiment"
-    WITH_EXPERIMENT = "with experiment"
 
 
 class PybammPipeline:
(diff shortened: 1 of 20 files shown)


pybop/parameters/parameter.py
```diff
@@ -763,22 +764,18 @@ def to_dict(self, values: str | ArrayLike | None = None) -> ParameterDict:
                 )
             return dict(zip(self._parameters.keys(), values_array, strict=False))
 
-    def to_pybamm_multiprocessing(self) -> list:
+    def to_inputs(self, values: np.ndarray | list[np.ndarray]) -> Inputs | list[Inputs]:
```
martinjrobins, line 767: is this signature correct? looks like the return type should be `list[ParameterDict]`

pybop/pipelines/_pybamm_pipeline.py
```diff
@@ -256,94 +234,75 @@ def build(self) -> None:
...
-    def solve(
-        self, calculate_sensitivities: bool = False
-    ) -> list[pybamm.Solution | FailedSolution]:
-        """
-        Run the simulation using the built model and solver.
+    def _solve_in_time_without_rebuild(
+        self, inputs: list[Inputs]
+    ) -> list[pybamm.Solution]:
+        """Solve in time without rebuilding the PyBaMM model."""
+        if len(inputs) == 1:
```
martinjrobins, line 241: do you need this, can you just pass in the list of length 1? I would hope if we (pybamm) did things correctly this should give the same results at the same speed :)

pybop/problems/base_problem.py
```diff
@@ -98,28 +73,104 @@ def sensitivity_analysis(self, n_samples: int = 256) -> dict:
...
+
+        Parameters
+        ----------
+        inputs : list[Inputs]
+            A list of input dictionaries.
+
+        Returns
+        -------
+        costs : np.ndarray
+            A 1D array of cost values with length (number of candidates).
```
martinjrobins, line 154: presumably number of candidates is len(inputs)? Might want to spell this out

martinjrobins (changes requested):
Thanks @NicolaCourtier, this looks great, just some minor stuff below

pybop/parameters/parameter.py
NicolaCourtier: Good point, it always return a list. Also I don't think we need `ParameterDict` now we only create `Inputs` dictionaries with single values. I've updated

PR #805, opened 2025-09-04
Generalise PybammCost to PybammOutputVariable

### LINKED ISSUE #804: Restructure: Can the cost functions have one input that they calculate from rather than two?
At the moment the cost functions deal with both variable name and data name, but this is very confusing - why not just have 'one' thing?  For example: .add_cost(pybop.costs.pybamm.MeanSquaredError("Voltage [V]", "Voltage [V]")). So, instead of having two "Voltage [V]", can we have one?

NicolaCourtier:
# Description

Rename the `PybammCost` class as `PybammOutputVariable`. Move fitting cost functionality out of the base class. Enforce a matching `signal` that exists in both the dataset and the model to avoid confusion (note that data columns can be renamed as part of the `pybop.Dataset` creation).

Fixes #804.

## Type of change

# Important checks:


martinjrobins (changes requested):
What is the reason for chainging `PybammCost` to `PybammVariable`? The filename is still called `cost`, and all the derived classes are still called `cost`, why not the base class? I don't have strong feelings either way, so happy to make the change if there is a good reason.

However, I don't think it is a good idea to force that the variable name and the data name are the same. We are now allowing *any* pybamm model, which will include models where the voltage has different names (e.g. "Voltage", or "V"). I'm happy to leave the `data_name` to be optional, in which case the default is the same as the `variable_name`

NicolaCourtier:
Hi @martinjrobins, thanks for the feedback.

The reason for changing `PybammCost` to `PybammVariable` (now `PybammOutputVariable`) is that this is more descriptive. As I understand it, this class enables the creation of any PyBaMM variable that can be calculated post-solution, including variables that are not single-value costs but also those which can be used for plotting e.g. the residual. I have updated the filename to `pybamm/output_variable.py`.

The reason for enforcing a single name for the signal/target variable is that this enables us to give each cost a meaningful name (again for plotting/validation later on). The place to match data names to model variable names is during the creation of the `Dataset`, which already allows complete flexibility in matching corresponding names. I can't see any reason to change names a second time and neither is any example of this included in the current restructure. An example of matching names via the dataset is included in `lgm50_pulse_validation.ipynb`.

(@Dibyendu-IITKGP any thoughts?)

martinjrobins:
Would you ever have the case where you have a Dataset, and you want to fit multiple models to it, which might have different variable names?

NicolaCourtier:
> Would you ever have the case where you have a Dataset, and you want to fit multiple models to it, which might have different variable names?

This doesn't seem particularly sensible? Not a common use case anyway
(but still achievable in a number of ways if really desired, e.g. by creating a duplicate variable in the dataset with the second name, updating the model variable name or creating duplicate output variables in the other models).

martinjrobins:
> This doesn't seem particularly sensible? Not a common use case anyway (but still achievable in a number of ways if really desired, e.g. by creating a duplicate variable in the dataset with the second name, updating the model variable name or creating duplicate output variables in the other models).

surely if you were developing your own model and wanted to compare the fit between your model and the standard pybamm spm model, this would be a common thing to do? I know I'm always comparing different models with each other for parameter inference purposes. But yes, there are workarounds so if you are sure that you don't want to specificaly enable this, then I'm happy to proceed with this PR

NicolaCourtier:
Yes, it's common to compare models but I think typically the models are attempting to fit/explain a particular dataset so all the models must predict the same signal (e.g. voltage), which can then be plotted for a visual comparison. (Of course, the other model variables that are not targets for fitting can be entirely different.)

Dibyendu-IITKGP:
> Hi @martinjrobins, thanks for the feedback.
> 
> The reason for changing `PybammCost` to `PybammVariable` (now `PybammOutputVariable`) is that this is more descriptive. As I understand it, this class enables the creation of any PyBaMM variable that can be calculated post-solution, including variables that are not single-value costs but also those which can be used for plotting e.g. the residual. I have updated the filename to `pybamm/output_variable.py`.
> 
> The reason for enforcing a single name for the signal/target variable is that this enables us to give each cost a meaningful name (again for plotting/validation later on). The place to match data names to model variable names is during the creation of the `Dataset`, which already allows complete flexibility in matching corresponding names. I can't see any reason to change names a second time and neither is any example of this included in the current restructure. An example of matching names via the dataset is included in `lgm50_pulse_validation.ipynb`.
> 
> (@Dibyendu-IITKGP any thoughts?)

I agree, it might create confusion for new users. Although, in the examples, we should clearly mention that while creating the dataset, user should rename the voltage section in such a way so as to match the voltage variable name in the model they are using.

NicolaCourtier:
> Yes, but there is no reason that a voltage variable in one model will have the same exact name as a voltage variable in another (e.g. could be "Voltage" in one and "V" in another). But I acknowledge that if we say "you can't do this" then it makes life easier on us ;) We just have to make sure we raise good errors, because I imagine it will be easy for users to mess this up and make a typo in the name of their voltage variable (I know I have many times in the past!).

Thanks @martinjrobins, I understand your use case better now and I think, as @Dibyendu-IITKGP suggests, it would be a great topic for an example. On balance, I think the advantages of having a shared name for new users and plotting outweighs the need to offer this for advanced users who can achieve the same result in a number of other ways (to be shown through examples).

martinjrobins (approved):
sounds good. Happy to merge this then, and we can add a suitable example later

PR #812, opened 2025-10-02
Pass parameters via ParameterValues class

NicolaCourtier:
# Description

Enables `pybop.Parameter` objects to be set directly in a `pybamm.ParameterValues` object to save the users having to first set dummy values for unknown parameters. Future updates should simplify this further so only a `pybop.Prior` (meaning prior information, not necessarily a prior distribution) is needed.

Also updates the method for scaling initial lithium concentrations for a changing geometry by updating the `parameter_values.set_initial_values` function.

## Type of change

# Important checks:


pybop/pybamm/eis_simulator.py
```diff
@@ -77,19 +74,23 @@ def __init__(
...
-        super().__init__(parameters=parameters)
-
         # Set-up model for EIS
         self._f_eval = f_eval
         model = self.set_up_for_eis(model)
         parameter_values = parameter_values or model.default_parameter_values
         parameter_values["Current function [A]"] = 0
 
+        # Unpack the uncertain parameters from the parameter values
+        parameters = Parameters()
```
martinjrobins, line 84: this could be done in the base simulator class so the code is not duplicated

benchmarks/benchmark_optim_construction.py
```diff
@@ -63,12 +47,27 @@ def setup(self, model, parameter_set, optimiser):
             }
         )
 
+        # Define fitting parameters
+        parameter_values.update(
+            {
+                "Negative electrode active material volume fraction": pybop.Parameter(
+                    "Negative electrode active material volume fraction",
```
martinjrobins, line 54: do we need to pass in the name in the __init__ anymore? Its a bit annoying to have to write it twice, and the name could be added when the simulator extracts the pybop parameters from the parameter_values class

martinjrobins (approved):
looks great, thanks @NicolaCourtier, just an optional suggestion to avoid having to write the name of the parameter twice:

benchmarks/benchmark_optim_construction.py
NicolaCourtier: Yes, this is what I was referring to. I reckon a follow-up PR could remove `parameter.name` completely, what do you think?

pybop/pybamm/eis_simulator.py
NicolaCourtier: I can do this in this PR.

martinjrobins: ok, no worries then, its only duplicated twice so not really much of a problem

PR #818, opened 2025-10-04
Bring example renaming from restructure

NicolaCourtier:
# Description

Apply the same re-naming as in the current `restructure` branch. Also remove the simple pulse fit example as before.

Note that the equivalent circuit notebooks were previously merged so, without this simplification, there are two notebooks (`ecm_hppc_identification` and `ecm_identification`) instead of one `ecm_pulse_identification`. We can merge them again later.

## Type of change

# Important checks:


brosaplanella (approved):
Looks good to me, thanks!

PR #839, opened 2025-11-24
Restructure priors

SarahRo:
# Description

This PR introduces a clearer distinction between priors and distributions used for sampling initial conditions.

## Renaming of classes and their methods and properties:
- ``BasePrior`` -> ``Distribution``
- ``JointPrior`` -> ``JointDistribution``
- ``Parameter.sample_from_prior`` -> ``Parameter.sample_from_distribution``
- ``Parameter._prior``-> ``Parameter._distribution``
- ``Parameter.prior`` -> ``Parameter.distribution``
- ``Parameters.sample_from_priors`` -> ``Parameters.sample_from_distributions``
- ``Parameters.priors ->Parameters.distributions``

## Changes to functionality:
- The ``pybop.Distribution`` object (formerly ``pybop.BasePrior``) is now constructed using a distribution of type ``scipy.stats.distributions.rv_frozen`` rather than ``scipy.stats.rv_continuous``. This means all parameters of the distribution are fixed by the user before passing it to the constructor of the class. The class no longer holds the values for the ``loc`` and ``scale`` paremeters. This was necessary to allow for distributions that have additional input parameters (e.g. ``scipy.stats.truncnorm``).
- A constructor for ``pybop.Distribution`` is now implemented. This allows a user to initialise a ``pybop.Distribution`` with any distribution of type ``scipy.stats.distributions.rv_frozen`` and use a finite difference approximation of the derivative of the log distribution.
- ``Distribution.mean`` is now a method rather than a property so that it matches the same method for ``scipy.stats.distributions.rv_frozen.``
- ``Distribution.sigma`` is replaced with  the method ``Distribution.std``, to match ``scipy.stats.distribtuions.rv_frozen``
- ``Distribution.support`` replace ``Distribution.bounds``, again to match ``scipy.stats.distributions.rv_frozen``.
- ``__repr__`` has been modified for ``pybop.Distribution`` and its subclasses to match information held by the class.
- ``pybop.Gaussian`` can now also be used for a truncated version of the normal distribution by providing the input ``truncated_at``. The underlying distribution in this case is ``scipy.stats.truncnorm``.
- A ``Parameter`` object can be initialised with either a ``pybop.Distribution`` or a ``scipy.stats.distribution.rv_frozen``. Previously, a ``Parameter`` required a ``BasePrior`` (now ``Distribution``).
- An error is thrown if bounds and a distribution are provided simultaneously to the constructor of ``Parameter``
- The set_bounds and remove_bounds methods have been removed from ``Parameter``
- The ``Parameters.update`` method has been changed to only accept initial_values and no bounds.

Fixes # (issue)

## Type of change

# Important checks:

### DIFF
--- a/pybop/parameters/distributions.py
+++ b/pybop/parameters/distributions.py
@@ -377,12 +377,12 @@ class Exponential(Distribution):
 
 class JointDistribution(Distribution):
     """
-    Represents a joint distribution distribution composed of multiple distribution distributions.
+    Represents a joint distribution composed of multiple distributions.
 
     Parameters
     ----------
     distributions : Distribution
-        One or more distribution distributions to combine into a joint distribution.
+        One or more distributions to combine into a joint distribution.
     """
 
     def __init__(self, *distributions: Distribution | stats.distributions.rv_frozen):
@@ -409,7 +409,7 @@ class JointDistribution(Distribution):
 
     def logpdf(self, x: float | np.ndarray) -> float:
         """
-        Evaluates the joint log-distribution distribution at a given point.
+        Evaluates the log of the joint distribution at a given point.
 
         Parameters
         ----------
@@ -434,7 +434,7 @@ class JointDistribution(Distribution):
 
     def logpdfS1(self, x: float | np.ndarray) -> tuple[float, np.ndarray]:
         """
-        Evaluates the first derivative of the joint log-distribution distribution at a given point.
+        Evaluates the first derivative of the log of the joint distribution at a given point.
 
         Parameters
         ----------


NicolaCourtier:
Thank you @SarahRo, this is definitely going in the right direction. Thanks for highlighting two potential issues, I would say:
1. `Distribution` can be moved out of the main namespace and become `pybop.parameters.Distribution` or similar. (Only needed if we keep `ParameterDistribution`...)
2. You've found a nice way of setting bounds which, I think, would allow us to move back to `pybop.Parameter(initial_value, distribution, bounds)`. As long as these options are passed together on construction, the result is consistent, but we will need to upgrade (or just remove) the post-set-up `set_bounds` and `remove_bounds` functions.

Also, please remove the `margin` attribute and the clip functions (hopefully no longer needed) and point the `Parameter.__call__` function at the `_initial_value` instead of the deprecated `_current_value`. Thanks!

pybop/parameters/distributions.py
```diff
@@ -315,32 +368,48 @@ def _dlogpdf_dx(self, x):
...
+            distribution
+            if isinstance(distribution, Distribution)
+            else Distribution(distribution)
+            for distribution in distributions
+        ]
 
     def logpdf(self, x: float | np.ndarray) -> float:
         """
-        Evaluates the joint log-prior distribution at a given point.
+        Evaluates the joint log-distribution distribution at a given point.
```
NicolaCourtier:
```suggestion
        Evaluates the log of the joint distribution at a given point.
```

PR #857, opened 2025-12-15
Align PyBaMM simulation methods

### LINKED ISSUE #783: Strategy pattern for `_pybamm_pipeline` post #780
### Feature description

With the addition of experiment support, the `_pybamm_pipeline` now houses a large amount of logic. It's handling model building, parameter management, solving strategy, threading, and state management in a single class. A strategy pattern should help split this and reduces the logic completed within the pipeline.

### Motivation

_No response_

### Possible implementation

Create `ExperimentStrategy` and `NonExperimentStrategy` classes following the below template.

```python
class Strategy():
    def __init__(...):

    def solve(self, inputs: List[Inputs]) -> List[pybamm.Solution]:
    
    def _solve_with_rebuild(self, inputs: List[Inputs]) -> List[pybamm.Solution]:
    
    def _solve_without_rebuild(self, inputs: List[Inputs]) -> List[pybamm.Solution]:

```

### Additional context

_No response_

NicolaCourtier:
# Description

Remove the custom build process for PyBaMM simulations without an experimental protocol. The aim is to improve maintainability by always using the build process provided by PyBaMM's Simulation class.

Also, change `batch_solve` to `solve_batch` to align with other function names such as `evaluate_batch`. Deprecate some of the warning/error catches that no longer appear to be needed.

Fixes #783 

## Type of change

# Important checks:

### DIFF
--- a/pybop/pybamm/simulator.py
+++ b/pybop/pybamm/simulator.py
@@ -87,6 +87,7 @@ class Simulator(BaseSimulator):
         for name, param in parameter_values.items():
             if isinstance(param, Parameter):
                 parameters.add(name, param)
+                self._parameter_values.update({name: "[input]"})
             elif isinstance(param, pybamm.InputParameter):
                 parameters.add(name, Parameter())
         super().__init__(parameters=parameters)
@@ -129,10 +130,8 @@ class Simulator(BaseSimulator):
         self.verbose = False
 
         # State
-        self._built_model = None
-        self._sim_experiment = None
+        self._simulation = None
         self._solve = None
-        self._calculate_sensitivities = False
 
         # Build
         self._input_parameter_names = self.parameters.names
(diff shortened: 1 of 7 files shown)


pybop/pybamm/simulator.py
```diff
@@ -253,28 +235,21 @@ def _set_up_solution_method(
...
             # Remove all voltage-based events when not using an experiment
             self._model.events = [e for e in self._model.events if "[V]" not in e.name]
 
-            # Build if only building once, otherwise build on evalution
-            if not self._requires_model_rebuild:
-                self.build_model()
-                self._solve = self._solve_in_time_without_rebuild
-            else:
-                self._solve = self._solve_in_time_with_rebuild
+        # Build if only building once, otherwise build on evaluation
```
brosaplanella, line 245: Not sure if it would save much computing time, but it might be better practice to store that as a property of the simulator, so it only gets computed once? Or do we expect that users could modify the simulator in a way that can change the need of rebuilding?

brosaplanella (approved):
Looks good, thanks! I made a small comment about setting the `_requires_model_rebuild` as a property, but feel free to ignore.

pybop/pybamm/simulator.py
NicolaCourtier: Thanks for the review! I'm sorry I don't quite follow this suggestion. This set-up is typically only performed once. The `self._simulation` (if any) is stored and the `self._solve` is pointed at the relevant solution method.

I think I will merge as it is, more updates can come later!

PR #858, opened 2025-12-15
Test lowest-direct dependencies

NicolaCourtier:
# Description

Add a continuous integration test that employs the lowest-direct dependencies, alongside the existing nightly check on the latest dependencies, to inform the lowest supported versions that we state in the `pyproject.toml`.

## Type of change

# Important checks:


NicolaCourtier:
Hi @agriyakhetarpal, I'm hoping you may have some advice!
The aim of this PR is to periodically check the lowest dependency versions that we say we support in the `pyproject.toml`. It wouldn't need to be run very often, but it would be nice to keep the lowest versions up-to-date. The command `uv pip install --resolution lowest-direct -e .[all,dev]` seems to do what I want locally, but simply adding this option to the [GitHub action](https://github.com/pybop-team/PyBOP/actions/runs/20235443725/job/58088801921) does not work. Would you recommend such a test and if so, how can it be implemented?

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -39,6 +39,40 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --resolution lowest-direct -e .[all,dev]
```
agriyakhetarpal:
```suggestion
      - uses: actions/checkout@v6
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-uv@v7
        with:
          python-version: ${{ matrix.python-version }}
      - name: Install dependencies
        run: |
          python -m uv pip install --resolution lowest-direct -e .[all,dev]
```

agriyakhetarpal (commented):
Hi @NicolaCourtier, I actually asked about this in the Scientific Python Discord server previously, but I forgot to respond to you here. Apologies! The command you have mentioned is the closest that will work for us here. I think the issue is that you are using `pip` instead of `uv pip`?

Besides this, one option is to adopt SPEC 0: https://scientific-python.org/specs/spec-0000/. It features a companion GitHub Action that we can use. However, PyBaMM has not adopted SPEC 0 yet, which is why I think it may not be the best idea for PyBOP. I think it would still be worth experimenting, however. See https://github.com/pybamm-team/PyBaMM/discussions/5202 for further discussion.

Another option I am aware of is to use a file like this: https://github.com/scikit-learn/scikit-learn/blob/ce1b377f3bb2580e7cc1e35ba7eed2131e3b8e8e/sklearn/_min_dependencies.py, though it will require us to add some extra CI setup.

NicolaCourtier:
Thanks very much for the suggestions @agriyakhetarpal

When trying the first, unfortunately it returns:
```Error: Unable to resolve action actions/setup-uv, repository not found```
See [here](https://github.com/pybop-team/PyBOP/actions/runs/21093641576/job/60668205535)

File: .github/workflows/nightly_dependency_tests.yaml [hunk 2]
```diff
@@ -39,6 +39,40 @@ jobs:
...
+        suite: ["unit", "integration", "examples", "notebooks"]
+
+    name: Test-${{ matrix.os }}-py-${{ matrix.python-version }}-${{ matrix.suite }})
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-uv@v7
+        with:
+          python-version: ${{ matrix.python-version }}
```
agriyakhetarpal, line 70:
```suggestion
      - name: Set up Python ${{ matrix.python-version }}
        uses: astral-sh/setup-uv@v7
        with:
          python-version: ${{ matrix.python-version }}
```

agriyakhetarpal (commented):
Ah, sorry, the action repository is hosted under Astral's organisation and not GitHub's.

NicolaCourtier:
Oops, didn't mean to close! Open again now, but there's a [new error](https://github.com/pybop-team/PyBOP/actions/runs/21102401587/job/60688412795)?

agriyakhetarpal:
Yes, this is because the `setup-uv` action installs a standalone `uv`, instead of adding a Python environment and installing `uv` from its Python distribution. Or so I think. Let's try!

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -39,6 +39,40 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m uv pip install --resolution lowest-direct -e .[all,dev]
```
agriyakhetarpal:
```suggestion
          uv pip install --resolution lowest-direct -e .[all,dev]
```

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -39,6 +39,40 @@ jobs:
...
+        suite: ["unit", "integration", "examples", "notebooks"]
+
+    name: Test-${{ matrix.os }}-py-${{ matrix.python-version }}-${{ matrix.suite }})
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: ${{ matrix.python-version }}
```
agriyakhetarpal, line 70:
```suggestion
          python-version: ${{ matrix.python-version }}
          activate-environment: "true"
```

agriyakhetarpal:
Triggered a run, and it looks like it's working: https://github.com/pybop-team/PyBOP/actions/runs/21102561524, though it fails because it's somehow trying to install `numpy==1.3.0` which is way too old. We've gone back to the Python 2 era :)

NicolaCourtier:
Great, then the test is working! It should fail for any minimum dependency that is too low. I only updated the numpy/scipy minimum in this PR! So we would expect the action to fail on develop. Thank you!

agriyakhetarpal (approved):
Awesome! Thanks, @NicolaCourtier!

NicolaCourtier:
Oh dear, [same error](https://github.com/pybop-team/PyBOP/actions/runs/21102801371/job/60689300462) after merging... looks like I can't just leave the numpy minimum unspecified (I assumed it would be be contemporary with the scipy version)

agriyakhetarpal:
Ah, yeah... I think a safe minimum for us would be NumPy 1.25.x series. I can't think of a reason we will use anything below that.

PR #860, opened 2025-12-15
Create parent Result class

NicolaCourtier:
# Description

Create a parent class for all types of optimisation and sampling result to clarify which attributes are shared.

Also, move `PosteriorSummary` attributes to the `SamplingResult` and deprecate the `pints.AdaptiveCovarianceMCMC` sampler.

## Type of change

# Important checks:


SarahRo (approved):
Overall this looks good to me and makes the code cleaner.

There is just one minor thing: I find it somewhat confusing where in the code the subclasses of Result are defined. OptimisationResult is defined in base_optimiser.py, but SamplingResult in mcmc_summary. Would it maybe make sense to define them in their own file?

PR #862, opened 2026-01-06
Allow multivariate parameters to be passed via pybamm.ParameterValues and define multivariate distributions in model space

### LINKED ISSUE #856: Allow MultivariateParameters to be set within ParameterValues
Align the set-up of `MultivariateParameters` with `Parameters` by allowing the user to mark the relevant parameters within a typical `pybamm.ParameterValues` object before passing it to a simulator. At the moment, the parameters attribute of a simulator has to be overridden after initialisation by `MultivariateParameters` and these new parameters are not currently shared when the simulator is passed to a problem - see [this example](https://github.com/pybop-team/PyBOP/blob/develop/examples/scripts/battery_parameterisation/bayesian_feature_fitting.py).

Note that, whenever two sets of parameters are joined together, any duplicate parameters need to be discarded. Other checks will be needed to ensure that if there is a multivariate distribution then there are no other optimisation parameters.

SarahRo:
# Description
The changes in the PR allow multivariate parameters to be passed via pybamm.ParameterValues. Moreover, multivariate distributions are now defined in the model space instead of the search space.

Introduces a MarginalDistribution class (subclass of Distribution) that is constructed from a BaseMultivariateDistribution. This is achieved by adding a method called 'marginal' to the multivariate distributions that returns the marginal distribution. The pybop.MarginalDistribution retains the original multivariate distribution in the parent_distribution attribute.

Introduces a MultivariateLogNormal class in order to define the distribution in the model space for the ep-bolfi example.

The MultivariateParameters class is removed. Instead, a check_multivariate method is added to pybop.Parameters() that checks whether a MarginalDistribution is present and if so whether all parameters have a MarginalDistribution with the same parent_distribution. This method is called in the constructor, the set method and the add method. The call to the method can be disabled in the add and set method by setting the parameter check_multivariate to false. This is, for example, used in the constructor in order to only call the method once all parameters have been added.

Fixes #856
 
## Type of change

# Important checks:


SarahRo:
> [!WARNING]
> The order of multivariate parameters is currently not verified inside pybop.Parameters(). They are assumed to be passed in order of the position property of the pybop.MarginalDistribution.

pybop/parameters/multivariate_distributions.py
```diff
@@ -154,14 +207,225 @@ class MultivariateGaussian(BaseMultivariateDistribution):
...
+        if isinstance(transformation, self.compatible_transformations) or (
+            isinstance(transformation, ComposedTransformation)
+            and all(
+                isinstance(trans, self.compatible_transformations)
+                for trans in transformation.transformations
+            )
+        ):
+            mean = transformation.to_search(self.properties["mean"])
+            covariance = transformation.convert_covariance_matrix(
+                self.properties["cov"], np.zeros(self._n_parameters)
```
NicolaCourtier:
```suggestion
                self.properties["cov"], mean
```
As far as I understand, the Jacobian should be evaluated at the transformed mean for this conversion.

NicolaCourtier (approved):
Looks great, thanks @SarahRo !

PR #867, opened 2026-01-19
Specify lowest requirements

NicolaCourtier:
# Description

This may be overkill... but here is a list of minimum requirements that passes the new continuous integration test.

Fixes `check_lowest_dependencies`.

## Type of change

# Important checks:


NicolaCourtier:
See the tests passing here: https://github.com/pybop-team/PyBOP/actions/runs/21141815763

pyproject.toml
```diff
@@ -46,28 +50,41 @@ docs = [
...
+    "nox[uv]>=2025",
+    "nbmake>=1.5.5",
+    "ipython>=8.38",  # for nbmake
+    "pyzmq>=27",  # for nbmake
+    "pre-commit>=4",
+    "pytest>=9",
+    "pytest-cov>=7",
+    "pytest-mock>=3.14",
+    "pytest-xdist>=3.7",
+    "ruff>=0.14",
```
agriyakhetarpal: This is generally a good idea, but I only realised now that we are checking for development dependencies in the test as well. I didn't realise it at the time in our previous discussion. We should drop that and only run tests with minimum dependency versions for `all`, and not `all,dev`.

pyproject.toml
```diff
@@ -46,28 +50,41 @@ docs = [
...
+]
 scifem = [
-  "scikit-fem>=8.1.0" # scikit-fem is a dependency for the multi-dimensional pybamm models
+    "scikit-fem>=8.1.0" # scikit-fem is a dependency for the multi-dimensional pybamm models
```
agriyakhetarpal: (Unrelated whitespace change)

```suggestion
   "scikit-fem>=8.1.0" # scikit-fem is a dependency for the multi-dimensional pybamm models
```

pyproject.toml
```diff
@@ -46,28 +50,41 @@ docs = [
...
+    "pyimpspec>=4",  # for ep-bolfi
+    "defusedxml>=0.6",  # for pyimpspec
+    "pillow>11",  # for pyimpspec
+    "xmlhelpy>=0.14",  # for ep-bolfi
+    "pyarrow>=14",  # for ep-bolfi
 ]
-
 pyprobe = [
-  "PyProBE-Data>=2.5.0;python_version>='3.11'"
+    "PyProBE-Data>=2.5.0;python_version>='3.11'"
```
agriyakhetarpal: (Unrelated whitespace change)

```suggestion
  "PyProBE-Data>=2.5.0;python_version>='3.11'"
```

pyproject.toml
```diff
@@ -46,28 +50,41 @@ docs = [
...
+    "h5py>=3.10",  # for ep-bolfi
+    "elfi>=0.8.6",  # for ep-bolfi
+    "numdifftools>=0.9.41",  # for elfi
+    "dask[distributed]>=2025.12",  # for elfi
+    "networkx>=3",  # for elfi
+    "pyimpspec>=4",  # for ep-bolfi
+    "defusedxml>=0.6",  # for pyimpspec
+    "pillow>11",  # for pyimpspec
+    "xmlhelpy>=0.14",  # for ep-bolfi
+    "pyarrow>=14",  # for ep-bolfi
```
agriyakhetarpal:
```suggestion
    "ep-bolfi>=3.0.2",
```

Cython is used only as a build-time dependency, so it shouldn't be specified as a runtime dependency. Similarly, `wheel` is only a CLI tool and we do not have much use for it in our runtime, so we should not list it. Is it a transitive dependency of us through a package? If yes, this is not good practice by them. We should drop both of them.

Similarly, I don't think it's helpful to specify the rest of the dependencies here, since we are a library rather than an application. In this case, doing this will increase the scope of our dependencies, making them non-transitive; i.e., if the package that depends on these projects stops doing so at some point, we may not get to know about it, and we will be specifying them for our users for no good reason.

pyproject.toml
```diff
@@ -46,28 +50,41 @@ docs = [
...
-    "pytest>=6",
-    "pytest-cov",
-    "pytest-mock",
-    "pytest-xdist",
-    "ruff",
-    ]
+    "nox[uv]>=2025",
+    "nbmake>=1.5.5",
+    "ipython>=8.38",  # for nbmake
+    "pyzmq>=27",  # for nbmake
```
agriyakhetarpal:
```suggestion
```

Same as https://github.com/pybop-team/PyBOP/pull/867#discussion_r2706378811

pyproject.toml
```diff
@@ -26,14 +26,18 @@ classifiers = [
 requires-python = ">=3.10, <3.13"
 dependencies = [
     "pybamm>=25.12.0",
-    "numpy",
-    "scipy>=1.12.0",
+    "numpy>=1.26",
+    "scipy>=1.12",
     "pints>=0.5.1",
+    "cma>=3",  # for pints
```
agriyakhetarpal:
```suggestion
```

Same as https://github.com/pybop-team/PyBOP/pull/867#discussion_r2706378811

pyproject.toml
```diff
@@ -26,14 +26,18 @@ classifiers = [
...
     "SALib>=1.5",
+    "multiprocess>=0.70.13",  # for SALib
+    "six>=1.16",  # for python-dateutil for pandas
+    "posthog>=3",  # for pybamm
```
agriyakhetarpal:
```suggestion
```

Same as https://github.com/pybop-team/PyBOP/pull/867#discussion_r2706378811

agriyakhetarpal:
Hi @NicolaCourtier, thank you for trying this out! I think it is okay to drop the lockfile if you think so. However, I am not convinced that explicitly specifying our transitive dependencies is a good approach.

Perhaps we should rethink our approach and instead use `uv pip compile` with `--lowest-direct`, instead of `install`. I am not sure what the best approach is, so we should probably look at how other libraries are managing their dependencies.

NicolaCourtier:
Thanks @agriyakhetarpal , these are good points! I agree there may be a better way. I didn't really understand what it would require when I suggested the test... still learning!

NicolaCourtier:
Ok, this seems better! All the Python 3.10 tests are passing: https://github.com/pybop-team/PyBOP/actions/runs/21218363394/job/61045936207 - I think we can ignore the 3.12 fails as off-topic for now.

(Locally I was testing the lowest requirements on 3.12 which I think misled me to add more lower pins than necessary.)

.github/workflows/nightly_dependency_tests.yaml
```diff
@@ -71,7 +71,8 @@ jobs:
           activate-environment: "true"
       - name: Install dependencies
         run: |
-          uv pip install --resolution lowest-direct -e .[all,dev]
+          uv pip install --resolution lowest-direct -e .[all]
+          uv pip install -e .[dev]
```
agriyakhetarpal:
```suggestion
          uv pip install -e .[dev]
          uv pip install --resolution lowest-direct -e .[all]
```

Nit: let's flip the order of these installation commands, as it is possible that some dependency might conflict or interact in funny ways – if we install the lowest-direct dependencies first, we can override the other packages in the environment.

agriyakhetarpal (approved):
Awesome, this is much better @NicolaCourtier! Thanks! :D

PR #870, opened 2026-01-21
Add currrent data processing methods

### LINKED ISSUE #869: Methods to infer a current function from data
## Background

Battery data is collected at varying time intervals depending on the experimental procedure (e.g. different time or voltage spacing). Charge throughput (or 'capacity') is usually measured alongside the current. All measurements are affected by a small amount of noise, which can depend on the equipment and set-up.

In PyBOP, we can simulate an experimental protocol based on measurement data. There are two issues I would like to address.

## 1) Interpolation

By default, when passing a set of current and time values, PyBaMM creates a linear interpolant to create a continuous input current. However, linear interpolation between measured points may not capture the true shape of the input current  and may not be in agreement with the measured throughput. PyBaMM offers pchip and cubic in addition to linear interpolation, but smoother functions are not necessarily closer to the true input. The current is often changed abruptly but only one data point is recorded just after the change, rather than both before and after the switch. The result is that the simulated throughput from a linearly interpolated current can be quite different from the measured throughput associated with an abrupt change.

PyBaMM issue 3446 is a request for zero-order hold interpolation, but it has not yet been implemented. While this technique would be able to simulate abrupt changes, the resulting input current may still not reproduce the measured charge throughput.

A better approximation may be obtained if we are able to use the charge throughput measurements to inform the shape of the input current between data points. One possible method would be to add extra points that ensure the integral of the linearly interpolated current to equal the measured throughput. For a given interval of time, an extra point may be needed if the magnitude of the charge throughput:
- is larger than the linear prediction
- is smaller than the linear prediction, but has the same sign as the change in current
- has the opposite sign to the change in current

In most cases, the noise level in the charge throughput should be low and so I would expect the extra points to be relatively accurate. For a large dataset, however, this process could result in a large number of extra points (up to one for every time interval), many of which provide very little extra information - which leads us onto the second issue.

## 2) Resolution

Battery datasets can be large and, for typical test protocols, it is often inefficient to reconstruct the input current from a large number of (slightly noisy) data points. To reduce the computational burden, I would suggest simplifying the input current where the variation is small (e.g. during constant-current segments, but not constant-voltage).

When deciding which sections can be safely downsampled, we need to check whether a given point is informative relative to its neighbours. Uninformative points may be removed or replaced to create constant sections where possible. Here is a possible approach.

If four or more consecutive current values are close (within a chosen tolerance) to the average of the first two values, the segment can be considered approximately constant. Any internal points (excluding two at each end) may be removed and the second and second-to-last current value may be adjusted to give a constant current which produces the same charge throughput as the original dataset.

## Approach

I would suggest adding a couple of "processing" methods which can be used to convert/generate new (consistent and/or downsampled) datasets from existing datasets. These methods should be entirely optional and designed to be used prior to setting up a simulator or optimisation problem.

NicolaCourtier:
# Description

1. Sort the existing utilities into relevant folders.

2. Add two methods for preparing datasets suitable for use as the current function in PyBaMM (more details in issue).

Fixes #869 

## Type of change

# Important checks:


pybop/processing/interpolate_current.py
```diff
@@ -0,0 +1,187 @@
...
+    except ValueError:
+        data_includes_throughput = False
+
+    # Iterative over neighbouring pairs of data points [i-1,i] and determine any sets of
+    # points that are uniformative and can be removed while keeping the same throughput
+    keep = np.full_like(time, True)
+    i = 1
+    offset = 0
+    while i + offset < time.shape[0] - 2:
+        delta_time = time[i] - time[i - 1]
```
brosaplanella: Should we move this to a later line, given that it is not used until the `if` statement in line 153?

pybop/processing/interpolate_current.py
```diff
@@ -0,0 +1,187 @@
...
+                # The measured throughput has a greater magnitude than possible with a hold,
+                # so add a point just after the first time to obtain the measured throughput
+                extra_times.append(time[i - 1] + tol)
+                extra_currents.append(
+                    (2 * delta_throughput + (current[i] - current[i - 1]) * tol)
+                    / delta_time
+                    - current[i]
+                )
+            elif switch_time > time[i] - tol:
+                # The measured throughput has a magitude of the opposite sign than possible
```
brosaplanella:
```suggestion
                # The measured throughput has a magnitude of the opposite sign than possible
```

brosaplanella: The sentence also reads a bit strange (maybe it's just me).

brosaplanella (approved):
Looks good! Just a couple of very minor comments

agriyakhetarpal:
I have self-requested a review here as I have a few comments on performance, and a few general-purpose improvements to typing and some subtle bugs – will put them in shortly!

NicolaCourtier:
Thanks @agriyakhetarpal, looking forward to your feedback! I just added a warning to the simulator, that was the last thing I wanted include here.

agriyakhetarpal (dismissed):
Hi @NicolaCourtier, thank you for waiting and apologies if I took too long! I actually opened this branch on my fork here: [agriyakhetarpal/pybop@`improving-nicola's-pr-gh-870`](https://github.com/agriyakhetarpal/PyBOP/tree/improving-nicola's-pr-gh-870). From it, I subsequently created #871 as a PR to your PR here, as I found it much easier to suggest these changes through a separate PR. Please feel free to look through them (I've added comments for some explanations), and rebase-merge that PR onto yours to bring those commits here when you are ready. Thanks!

PR #878, opened 2026-02-08
Use "Current" not "Current function" in datasets

NicolaCourtier:
# Description

Replaces "Current function [A]" with "Current [A]" as the name of the current data (not the input parameter), adds the `get_interpolant` functionality to the `Dataset` class and adds `import_pybamm_solution` as a shortcut for creating a dataset from a pybamm.Solution.

## Type of change

# Important checks:


pybop/processing/dataset.py
```diff
@@ -28,75 +30,37 @@ class Dataset:
 
     Parameters
     ----------
-    data_dictionary : dict or instance of pybamm.solvers.solution.Solution
+    data_dictionary : dict
         The experimental data to store within the dataset.
     domain : str, optional
         The domain of the dataset. Defaults to "Time [s]".
+    control_functions : list[str], optional
+        A list of function names for the control variables. Defaults to ["Current function [A]"].
```
SarahRo, line 38: Should this be ```["Current [A]"]```?

pybop/processing/dataset.py
```diff
@@ -28,75 +30,37 @@ class Dataset:
...
-            data_dictionary = data_dictionary.get_data_dict(variables=variables)
+        """Initialise a Dataset instance with data and a set of names."""
         if not isinstance(data_dictionary, dict):
-            raise TypeError(
-                "The input to pybop.Dataset must be a dictionary or a pybamm.Solution object."
-            )
+            raise TypeError("The input to pybop.Dataset must be a dictionary.")
         self.data = data_dictionary
         self.domain = domain or "Time [s]"
+        self.control_functions = control_functions or ["Current function [A]"]
```
SarahRo, line 52: Should this be ```["Current [A]"]```?

NicolaCourtier: So, the user needs (e.g.) `"Input function [unit]"` to exist as a parameter in their model and `"Input [unit]"` to exists in their dataset. By convention, the parameter has the extra string `" function"` (because it needs to be a function to run the model). My implementation takes the parameter name and later removes the extra string to get the data name. It could also be done the other way round, if that's clearer?
Unfortunately PyBaMM no longer seems to check if a parameter exists before updating it, so we would have to be a bit careful to prevent new, unused parameter names from being created.

SarahRo: Okay that makes sense. I would leave it as it is.

SarahRo (approved):

PR #880, opened 2026-02-08
perf: some changes while reviewing #870

agriyakhetarpal:
# Description

See #870. This PR is supposed to be merged into it. I shall add some comments inline here to explain the changes I have made. Please view the comments through https://github.com/pybop-team/PyBOP/pull/880/changes for a better view.

## Type of change

# Important checks:

### DIFF
--- a/pybop/processing/interpolate_current.py
+++ b/pybop/processing/interpolate_current.py
@@ -140,7 +140,7 @@ def downsample_constant_current(dataset: Dataset, tolerance: float = 1e-3) -> Da
     Q = cumulative_trapezoid(y=current, x=time, initial=0.0)
 
     # Iterative over neighbouring pairs of data points [i-1,i] and determine any sets of
-    # points that are uniformative and can be removed while keeping the same throughput
+    # points that are uninformative and can be removed while keeping the same throughput
     keep = np.full_like(time, True, dtype=bool)
     i = 1
     offset = 0
--- a/pybop/processing/interpolate_current.py
+++ b/pybop/processing/interpolate_current.py
@@ -134,7 +134,7 @@ def downsample_constant_current(dataset: Dataset, tolerance: float = 1e-3) -> Da
 
     # Iterative over neighbouring pairs of data points [i-1,i] and determine any sets of
     # points that are uniformative and can be removed while keeping the same throughput
-    keep = np.full_like(time, True)
+    keep = np.full_like(time, True, dtype=bool)
     i = 1
     offset = 0
     while i + offset < time.shape[0] - 2:
@@ -177,15 +177,15 @@ def downsample_constant_current(dataset: Dataset, tolerance: float = 1e-3) -> Da
         i += j + 2
         offset = 0
 
-    return pybop.Dataset(
+    return Dataset(
         {
-            "Time [s]": np.extract(keep, time),
-            "Current function [A]": np.extract(keep, current),
-            "Discharge capacity [A.h]": np.extract(keep, throughput) / 3600,
+            "Time [s]": time[keep],
+            "Current function [A]": current[keep],
+            "Discharge capacity [A.h]": throughput[keep] / 3600,
         }
         if data_includes_throughput
         else {
-            "Time [s]": np.extract(keep, time),
-            "Current function [A]": np.extract(keep, current),
+            "Time [s]": time[keep],
+            "Current function [A]": current[keep],
         }
     )
--- a/pybop/processing/interpolate_current.py
+++ b/pybop/processing/interpolate_current.py
@@ -106,7 +106,7 @@ def generate_consistent_current(dataset: Dataset, tolerance: float = 1e-3) -> Da
     )
 
 
-def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3):
+def downsample_constant_current(dataset: Dataset, tolerance: float = 1e-3) -> Dataset:
     """
     Generate a new dataset retaining only the informative points and consistency between the
     charge throughput and a linear interpolation of the current.
@@ -114,8 +114,8 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
     Arguments
     ---------
     dataset : pybop.Dataset
-        A dataset containg "Time [s]", "Current function [A]" and "Discharge capacity [A.h]".
-    delta_t : float
+        A dataset containing "Time [s]", "Current function [A]" and "Discharge capacity [A.h]".
+    tolerance : float
         A numerical tolerance in the units of current (A) used to determine if a data point
         is informative relative to its neighbours.
 
--- a/pybop/processing/interpolate_current.py
+++ b/pybop/processing/interpolate_current.py
@@ -92,16 +92,20 @@ def generate_consistent_current(dataset: Dataset, tolerance: float = 1e-3) -> Da
                 / 2
             )
 
+    extra_times = np.asarray(extra_times)
+    extra_currents = np.asarray(extra_currents)
+    extra_throughputs = np.asarray(extra_throughputs)
+
     time = np.concatenate((time, extra_times))
     current = np.concatenate((current, extra_currents))
     throughput = np.concatenate((throughput, extra_throughputs))
-    idx = np.argsort(time)
+    idx = np.argsort(time, kind="stable")
 
-    return pybop.Dataset(
+    return Dataset(
         {
-            "Time [s]": np.asarray(time)[idx],
-            "Current function [A]": np.asarray(current)[idx],
-            "Discharge capacity [A.h]": np.asarray(throughput)[idx] / 3600,
+            "Time [s]": time[idx],
+            "Current function [A]": current[idx],
+            "Discharge capacity [A.h]": throughput[idx] / 3600,
         }
     )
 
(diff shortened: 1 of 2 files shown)


pybop/processing/interpolate_current.py
```diff
@@ -1,21 +1,23 @@
+from __future__ import annotations
+
 import numpy as np
-from scipy.integrate import trapezoid
+from scipy.integrate import cumulative_trapezoid
 
-import pybop
```
agriyakhetarpal, line 4: I changed this, since it's better not to risk a circular import – but the better fix here (we can leave that for a later time) would be something like this:

```python
if TYPE_CHECKING:
    from .dataset import Dataset as DatasetType
```

but since we don't have type annotations for `pybop.Dataset`, that is a bit of a bigger task 😅

pybop/processing/interpolate_current.py
```diff
@@ -44,7 +46,11 @@ def generate_consistent_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
 
         tol = min(tolerance, delta_time / 2)
 
-        if linear_throughput == delta_throughput:
+        throughput_atol = tolerance * max(abs(current[i - 1]), abs(current[i]), 1.0)
+
+        if np.isclose(
+            linear_throughput, delta_throughput, rtol=0, atol=throughput_atol
+        ):
```
agriyakhetarpal: This was done to avoid comparing two floating-point numbers, since that's a no-no with a comparison operator. The previous condition would almost never hold for floats.

pybop/processing/interpolate_current.py
```diff
@@ -88,30 +94,34 @@ def generate_consistent_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
...
 
-    return pybop.Dataset(
+    return Dataset(
         {
-            "Time [s]": np.asarray(time)[idx],
-            "Current function [A]": np.asarray(current)[idx],
-            "Discharge capacity [A.h]": np.asarray(throughput)[idx] / 3600,
+            "Time [s]": time[idx],
+            "Current function [A]": current[idx],
+            "Discharge capacity [A.h]": throughput[idx] / 3600,
```
agriyakhetarpal, line 106: These `asarray` calls are redundant as they are already arrays

pybop/processing/interpolate_current.py
```diff
@@ -120,17 +130,20 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
     pybop.Dataset
         A new dataset containing the augmented time, current and charge throughput data.
     """
-    time = dataset["Time [s]"].copy()
-    current = dataset["Current function [A]"].copy()
+    time = np.asarray(dataset["Time [s]"])
+    current = np.asarray(dataset["Current function [A]"]).copy()  # we mutate this
```
agriyakhetarpal, line 130: We have avoided a copy here for `time`, but I had to keep the second one.

pybop/processing/interpolate_current.py
```diff
@@ -120,17 +130,20 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
...
+    except (ValueError, KeyError):
+        throughput = np.array([])  # just to keep type check happy
         data_includes_throughput = False
 
+    Q = cumulative_trapezoid(y=current, x=time, initial=0.0)
+
     # Iterative over neighbouring pairs of data points [i-1,i] and determine any sets of
-    # points that are uniformative and can be removed while keeping the same throughput
-    keep = np.full_like(time, True)
+    # points that are uninformative and can be removed while keeping the same throughput
```
agriyakhetarpal, line 141: Typo here, right? uniformative ➡️ uninformative?

pybop/processing/interpolate_current.py
```diff
@@ -120,17 +130,20 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
...
+        throughput: np.ndarray = dataset["Discharge capacity [A.h]"].copy() * 3600
         data_includes_throughput = True
-        throughput = dataset["Discharge capacity [A.h]"].copy() * 3600
-    except ValueError:
+    except (ValueError, KeyError):
+        throughput = np.array([])  # just to keep type check happy
         data_includes_throughput = False
 
+    Q = cumulative_trapezoid(y=current, x=time, initial=0.0)
+
```
agriyakhetarpal, line 139: Since we are calling `scipy.integrate.trapezoid()` on various slices, we can precompute this and get interval integrals in $O(1)$.

See below for the second part

pybop/processing/interpolate_current.py
```diff
@@ -154,17 +167,24 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
             # points and replace the second and second-to-last points with a constant current
             keep[i + 1 : i + j - 1] = False
             delta_time = time[i] - time[i - 1]
+
+            segment_integral = Q[i + j] - Q[i - 1]
             constant_current = (
-                2 * trapezoid(y=current[i - 1 : i + j + 1], x=time[i - 1 : i + j + 1])
+                2 * segment_integral
```
agriyakhetarpal, line 169: and this should become much closwr to $O(n)$, from the previous $O(n²)$ in the worst case, which was because of the above repeated trapezoid calls on each slice

pybop/processing/interpolate_current.py
```diff
@@ -173,15 +193,15 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
...
 
-    return pybop.Dataset(
+    return Dataset(
         {
-            "Time [s]": np.extract(keep, time),
-            "Current function [A]": np.extract(keep, current),
-            "Discharge capacity [A.h]": np.extract(keep, throughput) / 3600,
+            "Time [s]": time[keep],
+            "Current function [A]": current[keep],
+            "Discharge capacity [A.h]": throughput[keep] / 3600,
```
agriyakhetarpal, line 196: When the condition is a boolean, as is the case here, you can simply replace it 😉 see the second half of the examples in https://numpy.org/devdocs/reference/generated/numpy.extract.html

pybop/simulators/failed_solution.py
```diff
@@ -17,7 +17,7 @@ class FailedVariable:
 
     name: str
     data: np.ndarray = field(default_factory=lambda: np.asarray([np.inf]))
-    sensitivities: {str, np.ndarray} = field(default_factory=dict)
+    sensitivities: dict[str, np.ndarray] = field(default_factory=dict)
```
agriyakhetarpal, line 20: This annotation was incorrect because `{}` is for defining sets, and we can't do that for type hints. I've changed it for `self._variables` below as well.

pybop/processing/interpolate_current.py
```diff
@@ -154,17 +167,24 @@ def downsample_constant_current(dataset: pybop.Dataset, tolerance: float = 1e-3)
...
             constant_current = (
-                2 * trapezoid(y=current[i - 1 : i + j + 1], x=time[i - 1 : i + j + 1])
+                2 * segment_integral
                 - current[i - 1] * delta_time
                 - current[i + j] * (time[i + j] - time[i + j - 1])
             ) / (time[i + j] + time[i + j - 1] - time[i] - time[i - 1])
+
+            old_current_i = current[i]
+            old_current_j = current[i + j - 1]
+
```
agriyakhetarpal, line 176: So this is what I think was a bug here (?). In `downsample_constant_current` we computed `constant_current`, but we immediately assigned them like this:

```python
current[i] = constant_current
current[i + j - 1] = constant_current
```

but in the `data_includes_throughput` code path (where we are doing the throughput correction), we had `constant_current - current[i]` and `(constant_current - current[i + j - 1])`, which would then both be zero, as we overwrote them and nothing gets corrected?

I've simply cached these values into `old_current_i` and `old_current_j` here. Let me know what you think!

pybop/processing/interpolate_current.py
NicolaCourtier: Thanks! (There's a few other places we should do this...)

pybop/processing/interpolate_current.py
NicolaCourtier: I agree this is better but I think we can use a fixed tolerance here (e.g. 1e-10) on the value, which is in As.

pybop/processing/interpolate_current.py
NicolaCourtier: Nice improvement

pybop/processing/interpolate_current.py
NicolaCourtier: Yep! Thanks for fixing my typos

pybop/processing/interpolate_current.py
NicolaCourtier: Thanks, great catch! You are right. These lines were technically tested but the change in throughput is very small so it wasn't detected.

NicolaCourtier (approved):
Thanks for these improvements and for explaining them, very helpful.

pybop/processing/interpolate_current.py
agriyakhetarpal: Makes sense to me – will change!

pybop/processing/interpolate_current.py
agriyakhetarpal: Agreed, we should hunt them down. We don't have Mypy running either, and I assume it will flag a lot of errors already.

agriyakhetarpal:
I changed to a fixed tolerance as you suggested, in 5093fc6e123d5bae8a5d5ba39b6d45cc2c90505e. I'll merge this and bring the changes to your branch (GitHub is conveniently down right now 😅, but I ran the test suite locally, which is green).

PR #883, opened 2026-02-10
Update scientific-python/issue-from-pytest-log-action to v1.5.0

agriyakhetarpal:
# Description

See https://github.com/pybop-team/PyBOP/pull/879#issuecomment-3878252269

cc: @NicolaCourtier

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybop-team/PyBOP/blob/develop/CHANGELOG.md) to document the change (include PR #).

# Important checks:

Please confirm the following before marking the PR as ready for review:
- No style issues: `$ pre-commit run` or `$ nox -s pre-commit` (see [CONTRIBUTING.md](https://github.com/pybop-team/PyBOP/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- All tests pass: `nox -s tests`
- The documentation builds: `nox -s doctest`
- Code is commented for hard-to-understand areas
- Tests added that prove fix is effective or that feature works

PR #879, opened 2026-02-08
Open issues if PyBaMM dev and lowest-dependency tests fail

agriyakhetarpal:
# Description

This PR closes #297; it adds https://github.com/scientific-python/issue-from-pytest-log-action in order to allow opening issues from failures reported in our scheduled nightly tests. I've also bumped a number of our GitHub Actions versions, which were getting outdated.

## Type of change

Please add a line in the relevant section of [CHANGELOG.md](https://github.com/pybop-team/PyBOP/blob/develop/CHANGELOG.md) to document the change (include PR #).

# Important checks:

Please confirm the following before marking the PR as ready for review:
- No style issues: `$ pre-commit run` or `$ nox -s pre-commit` (see [CONTRIBUTING.md](https://github.com/pybop-team/PyBOP/blob/develop/CONTRIBUTING.md#installing-and-using-pre-commit) for how to set this up to run automatically when committing locally, in just two lines of code)
- All tests pass: `nox -s tests`
- The documentation builds: `nox -s doctest`
- Code is commented for hard-to-understand areas
- Tests added that prove fix is effective or that feature works

PR #884, opened 2026-02-10
Fix nightly workflow again

agriyakhetarpal:
# Description

From #879 and #883: TIL that `uv pip` does not accept `-y` for "yes", it installs/uninstalls without prompting the user for input. This was missed when I changed `pip` to `uv pip` in the "latest" nightly tests. 

## Type of change

# Important checks:


agriyakhetarpal:
Side note: we shouldn't install the `dev` dependencies alongside `all`. Instead, we should use the newer dependency groups. I'll prepare a PR for this (edit: did that with #885).

NicolaCourtier (approved):

PR #885, opened 2026-02-10
Use dependency groups

agriyakhetarpal:
# Description

See https://packaging.python.org/en/latest/specifications/dependency-groups/ for docs; they offer better ways for us to install development or documentation dependencies without the added burden of exposing them to our users through our PyPI metadata.

## Type of change

# Important checks:


agriyakhetarpal:
Looks like Read The Docs is not running here. Are its PR previews disabled? If I remember correctly, they used to be enabled a long time back.

.readthedocs.yaml
```diff
@@ -3,7 +3,12 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.11"
+    python: "3.13"
+  jobs:
+    # https://docs.readthedocs.com/platform/latest/build-customization.html#install-dependencies-from-dependency-groups
+    install:
+      - pip install --upgrade pip
+      - pip install --group 'docs'
```
NicolaCourtier: Should it be `pip install --group docs`  or are the `''` needed here?

noxfile.py
```diff
@@ -10,19 +10,22 @@
 # Environment variables to control CI behaviour for nox sessions
 PYBOP_SCHEDULED = int(os.environ.get("PYBOP_SCHEDULED", 0))
 PYBAMM_VERSION = os.environ.get("PYBAMM_VERSION", None)
+PYPRPOJECT = nox.project.load_toml("pyproject.toml")
```
NicolaCourtier: Typo? `PYPROJECT`

.readthedocs.yaml
agriyakhetarpal: I think we can drop it; the docs used the quotes but we don't necessarily need to follow them :D

noxfile.py
agriyakhetarpal: Oops, thanks!

readthedocs.yaml
agriyakhetarpal, line 1: It turns out that we had two RTD configuration files. The correct file to use is `.readthedocs.yaml`, so this one was never used anyway!

NicolaCourtier (approved):
Looks good! Thanks @agriyakhetarpal

agriyakhetarpal:
All green! Merging :D

PR #889, opened 2026-02-13
Update set_initial_state

NicolaCourtier:
# Description

Update the `set_initial_state` functions for models that can take a initial voltage and determine the initial state of charge or stoichiometry. Also, move the `Interpolant` and `InverseOCV` classes into a new model `utils.py`.

## Type of change

# Important checks:


brosaplanella (commented):
I only have one comment: there's quite a lot of code duplication in `set_initial_state` (as far as I can tell). It might be better to have a `pybop.lithium_ion.BaseModel` with that method, from which all other classes are initialised.

I'm not too attached to this though, so if you think it's better to keep it as is, happy to approve.

pybop/models/lithium_ion/base_model.py
```diff
@@ -0,0 +1,127 @@
...
+from pybamm import (
+    FunctionParameter,
+    Parameter,
+)
+from pybamm import lithium_ion as pybamm_lithium_ion
+
+from pybop.models.lithium_ion.utils import InverseOCV
+
+
+class BaseModel(pybamm_lithium_ion.BaseModel):
```
NicolaCourtier: Please can we name this class `BaseGroupedModel` or similar? (To avoid confusion with the other `BaseModel` classes)

PR #890, opened 2026-02-15
Fix parameter assignment

NicolaCourtier:
# Description

Fix the assignment of parameters to problems, when the problems within a `MetaProblem` depend on a different set of parameters. Also, allow passing a dictionary of parameter names and values to `BaseSimulator` and update the corresponding test.

## Type of change

# Important checks:


NicolaCourtier:
:warning: More care is needed regarding the order of the parameters to ensure that each problem returns the correct sensitivities (gradient in the cost with respect to each parameter). We use an `OrderedDict` in the `Parameters` class to preserve the parameter order. Then, when we construct a `Problem`, we first add the parameters from the simulator, followed by any extra ones for the cost. Duplicate parameters (those with the same parameter name) are ignored and a warning is thrown. But when we combine problems into a `MetaProblem`, we do not know if the problems have the same optimisation parameters and in which order they may be. So, this fix for the `MetaProblem` does not currently work for gradient-based methods if the optimisation parameters differ between the two problems.

SarahRo (commented):
The changes look good to me.

One thing I noticed is that the parameter assignment in MetaProblem does not seem to be tested as in there is no test that would fail without this bug-fix.

SarahRo (approved):

PR #911, opened 2026-03-10
fix voronoi plot

### LINKED ISSUE #910: [Bug]: Surface plot and contour plot do not agree in energy_based_electrode_design.ipynb
### Python Version

3.13, 3.12

### Describe the bug

The surface and contour plot for the result in energy_based_electrode_design.ipynb should show similar results but they differ significantly.

In energy_based_electrode_design.ipynb, add line to produce contour plot: 
`result.plot_contour()`
`result.plot_surface();`

### Relevant log output

```shell

```

SarahRo:
# Description
The surface plot in energy_based_electrode_design.ipynb should be closer to the contour plot of the result.

Cause of the difference is that the cost is recomputed for plotting for the contour plot whereas the surface plot uses `result.cost`. `result.cost` is identical to `logger.cost_convergence` and hence, rather than storing the cost for the corresponding points `result.x_model`, it contains the best cost up to each point in the optimisation process. To fix this, result.cost now returns the same as logger.cost and result.cost_convergence replaces result.cost.

Fixes #910 

## Type of change

# Important checks:


NicolaCourtier (approved):
Great! Thanks @SarahRo

PR #925, opened 2026-04-02
Transform distributions

### LINKED ISSUE #671: Align the setting and passing of bounds
We should perform a thorough check of everywhere bounds are passed, add tests for different formats and update the documentation. The application of the bounds in BaseOptimiser was changed in #661 and not all bases were covered. Things to consider:
- bounds in different formats e.g. lists and dictionaries
- inappropriate formats such as strings
- one-sided bounds
- infinite bounds, including transformed bounds e.g. log(0)
- when to remove bounds e.g. via the input bounds=None
- bounds for plotting

NicolaCourtier:
# Description

The implementation of parameters, distributions and transformations is over-complicated. The aim here is to simplify the definitions and interactions between these classes to improve maintainability. A new `UnboundedDistribution` is created to define the behaviour of a parameter without a statistical distribution. The `get_transformed_distribution` functionality is added to both univariate and multivariate distributions.

## Type of change

# Important checks:


pybop/parameters/distributions.py
```diff
@@ -15,16 +22,25 @@ class Distribution:
...
     distribution : scipy.stats.distributions.rv_frozen
         The underlying continuous random variable distribution.
+    n_parameters : int
+        The number of dimensions (default: 1).
     """
 
     def __init__(
         self,
         distribution: stats.distributions.rv_frozen | None = None,
+        properties: dict | None = None,
```
martinjrobins: need to add this to the docstring

pybop/parameters/distributions.py
```diff
@@ -15,16 +22,25 @@ class Distribution:
     ----------
     distribution : scipy.stats.distributions.rv_frozen
         The underlying continuous random variable distribution.
+    n_parameters : int
+        The number of dimensions (default: 1).
     """
 
     def __init__(
         self,
         distribution: stats.distributions.rv_frozen | None = None,
```
martinjrobins: what does distribution=None mean? unbounded? need to add to the docstring

pybop/parameters/parameter.py
```diff
@@ -112,120 +70,78 @@ class Parameter:
 
     def __init__(
         self,
-        distribution: stats.distributions.rv_frozen | Distribution | None = None,
+        distribution: Distribution | None = None,
```
martinjrobins: again, what is distribution=None mean?

martinjrobins (changes requested):
looks good, couple of questions around distribution = None / unbounded

pybop/parameters/distributions.py
NicolaCourtier: To make things clearer, I have added the `BaseDistribution` class - which can be used as a placeholder for an empty parameter.

pybop/parameters/parameter.py
NicolaCourtier: Added to the description:
```
distribution : BaseDistribution, optional
        Probability distribution for the parameter. If None, an empty
        `pybop.BaseDistribution` will return a NotImplementedError for any
        functionality that requires a distribution, such as `rvs`.
```

martinjrobins: but what does it mean from a user perspective? Does it mean that the parameter is unbounded? Or that it can't be used in sampling, or multi-start optimisation?

NicolaCourtier: If no distribution, bounds or initial value is specified, an empty `Parameter` with an empty `BaseDistribution` will be created. It doesn't really mean anything... but it should return a somewhat helpful error when something is required but not implemented. It has no bounds and yes, it cannot be used for sampling (no `mean` or `cov`) or multi-start optimisation (no `rvs`).

martinjrobins (approved):

PR #928, opened 2026-04-02
Allow user to choose plotting library

### LINKED ISSUE #855: Allow users to choose plotting library
### Feature description

Allow users to choose between plotly and matplotlib by separating the creation of plotting data and options from the creation of the plot.

### Motivation

To enable users to create plots in their preferred style and suitable for publication, e.g. using the [`scienceplots` package](https://github.com/garrettj403/SciencePlots).

### Possible implementation

It may be possible to extend our `StandardPlot` for different plotting libraries, or we could change the plotting class structure to make it more flexible.

### Additional context

The Voronoi functions are based on the [`pints.plot.surface`](https://github.com/pints-team/pints/blob/main/pints/plot/_surface.py). For comparison, Matplotlib has a [`voronoi_plot_2d`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.voronoi_plot_2d.html) function.

Where possible, we would like to follow similar design principles to PyBaMM. For example, our `result.plot_convergence` allows users to view their results in a similar way to PyBaMM's [`solution.quick_plot`](https://github.com/pybamm-team/PyBaMM/blob/develop/src/pybamm/plotting/quick_plot.py).

SarahRo:
# Description

This PR adds the following features:

- All plotting functions can now be used with either `plotly` or `matplotlib.pyplot` as the plotting backend.
- The backend can be chosen by 
    - calling `pybop.plot.use_backend('plotly')'
    - passing the backend as a keyword argument `backend = ...` to the plotting function.
    - creating a figure with either plotting library and passing it to the plotting function in the 'figures' keyword argument.
- Figures and axes can be passed as keyword arguments to plotting functions to allow full user control over figure layout and positioning within subplots.

Fixes #855 

## Type of change

# Important checks:


pybop/plot/convergence.py
```diff
@@ -33,22 +39,40 @@ def convergence(result: "Result", show=True, **layout_kwargs):
...
-    if show:
-        fig.show()
+    backend.legend(fig, axes=ax)
 
-    return fig
+    # Display or return figure
+    if show:
+        backend.show_figure(fig)
+    else:
+        return fig
```
NicolaCourtier, line 79: Is there a reason why this function only returns the figure if show=False?
Same for plot.samples and plot.distribution, but not the others?

pybop/plot/parameters.py
```diff
@@ -2,12 +2,15 @@
...
 from pybop.plot.standard_plots import StandardSubplot
+from pybop.plot.util import get_backend_from_figure
 
 if TYPE_CHECKING:
     from pybop._result import Result
 
 
-def parameters(result: "Result", show=True, **layout_kwargs):
+def parameters(
+    result: "Result", show: bool = True, backend: str = None, figures=None, axes=None
```
NicolaCourtier: Add optional `title` to match the others?

pybop/plot/trajectories.py
```diff
@@ -0,0 +1,69 @@
...
+    yaxis_title: str = None,
+    labels=None,
+    label_width=20,
+    show=True,
+    backend=None,
+    figures=None,
+    axes=None,
+):
+    """
+    Quickly plot one or more trajectories using Plotly.
```
NicolaCourtier:
```suggestion
    Quickly plot one or more trajectories.
```

pybop/plot/samples.py
```diff
@@ -1,109 +1,140 @@
...
+    # Process input
+    figures, axes, create_figure, single_axis = backend.parse_input_axes(
+        figures, axes, num_plots=result.n_parameters
     )
-    fig.update_layout(**kwargs)
-    fig.show()
+
+    for i in range(result.n_parameters):
+        ax = axes[i % len(axes)]
+        title = "Parameter Trace Plot" if single_axis else f"Parameter {i} Trace Plot"
```
NicolaCourtier: Suggest using the name of each parameter instead of numbering (here and below)

NicolaCourtier (commented):
This is excellent, thanks @SarahRo !

Some minor comments and questions:
- do you think we should deprecate `StandardPlot` and `StandardSubplot` in favour of the standardised backend functionality, since they only seem to be used once each?
- there seems to be an issue with the Voronoi plot in `optimising_with_scipy_minimize.py` although it works elsewhere, please could you check if this is due to changes made in this PR?

pybop/plot/convergence.py
SarahRo: Returning the figure for show=True makes sense for plotly but not matplotlib. plt.show() will clear the figure and it is therefore no longer useful.

SarahRo:
> This is excellent, thanks @SarahRo !
> 
> Some minor comments and questions:
> 
>     * do you think we should deprecate `StandardPlot` and `StandardSubplot` in favour of the standardised backend functionality, since they only seem to be used once each?
> 
>     * there seems to be an issue with the Voronoi plot in `optimising_with_scipy_minimize.py` although it works elsewhere, please could you check if this is due to changes made in this PR?

I'm looking into this. I simplified the voronoi plot when replicating it in matplotlib so that is likely what is causing the issue. It must be a bit of an edge case since it's working for most examples.

I will also look into the changes required for deprecating `StandardPlot` and `StandardSubplot`.

NicolaCourtier (approved):
Looks great, thanks @SarahRo !

PR #938, opened 2026-04-10
Rename Sobol sensitivity method

### LINKED ISSUE #657: [Bug]: Global Sensitivities struggle with non-bounded parameters
### Python Version

3.12

### Describe the bug

Since our non-bounded parameters still have 'bounds' see here, sampling methods that use these parameter bounds can result in numerical instabilities (for example, physical parameters such as electrode thickness can have unphysical negative bounds). Due to this issue, it's very much a possibility that the sampling methods can result in `nan` if they experience this issue.

Run the current `maximum_likelihood.py` example with `compute_sensitivities` optimiser arg, which will result in `nan` values for the sensitivities calculation.

### Relevant log output

```shell

```

NicolaCourtier:
# Description

Add a check for finite bounds and specify the method by renaming `sensitivity_analysis` to `get_sobol_sensitivities`.

Fixes #657

## Type of change

# Important checks:


NicolaCourtier:
Alternatively, we can deprecate the function and move SALib into the optional dependencies, as SALib is only used directly in the examples. What do you think @Dibyendu-IITKGP ?

pybop/analysis/sensitivity_analysis.py
```diff
@@ -36,9 +37,13 @@ def sensitivity_analysis(
     Sensitivities : dict
     """
 
+    bounds_array = problem.parameters.get_bounds_array()
+    if not np.isfinite(bounds_array).all():
```
agriyakhetarpal, line 41:
```suggestion
    if np.any(~np.isfinite(bounds_array)):
```

nit: minor optimisation where we short-circuit as soon as it hits the first `True`

agriyakhetarpal (approved):
Neat, thanks, I see that this was removed completely! Let's also update this section as well: https://pybop-docs.readthedocs.io/en/latest/installation.html#optional-dependencies

NicolaCourtier:
Thanks @agriyakhetarpal, I've added the more recent optional dependencies to the docs, good spot!

PR #945, opened 2026-05-26
Add saving/loading example

NicolaCourtier:
# Description

Add an example script that shows how to save and load an optimisation result.

## Type of change

# Important checks:

### DIFF
--- a/examples/scripts/getting_started/saving_and_loading_results.py
+++ b/examples/scripts/getting_started/saving_and_loading_results.py
@@ -64,12 +64,17 @@ result.save(save_path + "saved_result_object.pkl")
 result.save_data(save_path + "saved_result_data.json", to_format="json")
 
 # Load the result
-result = pybop.Result.load(save_path + "saved_result_object.pkl")
-result = pybop.Result.load_data(
+result_from_pkl = pybop.Result.load(save_path + "saved_result_object.pkl")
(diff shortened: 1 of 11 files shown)


examples/scripts/getting_started/saving_and_loading_results.py
```diff
@@ -0,0 +1,75 @@
...
+os.makedirs(os.path.dirname(save_path), exist_ok=True)
+result.save(save_path + "saved_result_object.pkl")
+result.save_data(save_path + "saved_result_data.json", to_format="json")
+
+# Load the result
+result = pybop.Result.load(save_path + "saved_result_object.pkl")
+result = pybop.Result.load_data(
+    save_path + "saved_result_data.json", file_format="json"
+)
+
```
SarahRo, line 71: Would it make sense to have two Result objects here, one for the result loaded from the pickle file and one for the result reconstructed from the data in the json file? Maybe even plot both to illustrate that they hold the same data?

It might otherwise look like the load_data function is needed to populate the result object loaded from the pickle file.

NicolaCourtier: Sure, either is fine by me

brosaplanella (approved):
I agree with Sarah's comment on the two objects, but otherwise looks good!

